Issue #62, opened 2015-04-27
Wrong cost when usage is 0


eloycoto:
Morning, 

I've realised that when the usage is 0 in cdrc, the cost is -1. I tried to debug but I can't found the problem. 

I'm using the following csv: 

```csv
32b2bc20-ecca-11e4-bf65-0800272f1d8c,1001,1002,*out,call,2015-04-27 10:43:10,2015-04-27 10:43:10,0,cgrates.org,*voice
762b6920-ecca-11e4-84a3-0800272f1d8c,1001,1002,*out,call,2015-04-27 10:45:03,,0,cgrates.org,*voice
```
This is the CDRC client section: 

```js
"cdrc": {
      "*default": {
          "enabled": true,
          "cdrs": "internal",
          "cdr_format": "csv",
          "field_separator": ",",
          "run_delay": 0,
          "data_usage_multiply_factor": 1024,
          "cdr_in_dir": "/var/log/cgrates/cdrc/in",
          "cdr_out_dir": "/var/log/cgrates/cdrc/out",
          "cdr_source_id": "freeswitch_csv",
          "cdr_filter": "",
          "cdr_fields":[
              {"tag": "tor", "cdr_field_id": "tor", "type": "cdrfield", "value": "9", "mandatory": true},
              {"tag": "accid", "cdr_field_id": "accid", "type": "cdrfield", "value": "0", "mandatory": true},
              {"tag": "reqtype", "cdr_field_id": "reqtype", "type": "cdrfield", "value": "4", "mandatory": true},
              {"tag": "direction", "cdr_field_id": "direction", "type": "cdrfield", "value": "3", "mandatory": true},
              {"tag": "tenant", "cdr_field_id": "tenant", "type": "cdrfield", "value": "8", "mandatory": true},
              {"tag": "category", "cdr_field_id": "category", "type": "cdrfield", "value": "4", "mandatory": true},
              {"tag": "account", "cdr_field_id": "account", "type": "cdrfield", "value": "1", "mandatory": true},
              {"tag": "subject", "cdr_field_id": "subject", "type": "cdrfield", "value": "1", "mandatory": true},
              {"tag": "destination", "cdr_field_id": "destination", "type": "cdrfield", "value": "2", "mandatory": true},
              {"tag": "setup_time", "cdr_field_id": "setup_time", "type": "cdrfield", "value": "5", "mandatory": true},
[...]

PR #63, opened 2015-04-27
Issue #62: Fix the cost issue when duration is 0.

eloycoto:
This fix the issue #62, In this case we create a new callCost struct, and if the duration is 0, we return the default cost.

If duration is bigger than 0 the system process the call as normal. 

I tried to make a test, but I couldn't :-(. It doesn't exist any example of how to create a new CDRServer in the tests, and I can't create one, I tried with multiple options but I always fail. 

If you show me an example of new CdrServer on one test, I'll provide the test for this function. 

Regards

### DIFF
--- a/engine/cdrs.go
+++ b/engine/cdrs.go
@@ -199,10 +199,13 @@ func (self *CdrServer) getCostsFromDB(cgrid, runId string) (cc *CallCost, err er
 
 // Retrive the cost from engine
 func (self *CdrServer) getCostFromRater(storedCdr *StoredCdr) (*CallCost, error) {
-	if storedCdr.Usage == time.Duration(0) { // failed call,  nil cost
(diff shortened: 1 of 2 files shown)


danbogos:
Hi Eloy,

Thanks for this.

Unfortunately this fix only handles rated/postpaid CDRs and not addressing the prepaid/pseudoprepaid ones. We will need a fix in the calldesc.go, so will let you know when that done

eloycoto:
Hi Dan, 

A quick one, in this case this duration check was in this function(Before my change), if you disable duration == 0 and process the call as normal with CallDescriptor, you get the good result too.

The if was added in the commit 7677a5ff89526756866cbee54c7dd31ac067b310. I can't figure why this duration check was added, but I think that it was for performance.. but no idea 

Hope that this help. 
Regards.

PR #113, opened 2015-07-08
Storage: Added jsonb support in cdrs_extra.extra_fields

eloycoto:
Hi, 

Extra_info are now saving into PostgreSQL database as text. PostgreSQL 9.4+ support native json support, so it's easy to make specified queries in the database using json. 

No changes in the code, Jsonb queries are the same as a text. 

About Jsonb in PostgreSQL:
- Datatypes: http://www.postgresql.org/docs/9.4/static/datatype-json.html
- Functions: http://www.postgresql.org/docs/9.4/static/functions-json.html

PS: PostgreSQL version in Debian stable is 9.4


danbogos:
Thanks!

PR #117, opened 2015-07-13
Storage: Added jsonb support in cost_details.timespans

eloycoto:
Hi, 

Related with PR113, the same but in this case for cost_details. 

Thanks


danbogos:
Thanks!

PR #309, opened 2015-11-30
Detect active single-leg channels

bhepp:
Only bridged calls will have a non-empty `call_uuid` in `show channels`. There are some valid single-leg scenarios (conference calls for example), which are not billed properly at the moment, because such calls are incorrectly detected as inactive:
```
CGRateS[8509]: <SM-FreeSWITCH> Sync active channels, stale session detected, uuid: 3030f113-1d83-4e80-b304-7e7c50084b7e
```

This PR fixes it by checking the `uuid` variable in case the `call_uuid` is empty. If the call leg was to be bridged, its `call_uuid` will be set to the `uuid` anyway. They are not always equivalent though, that's why it has to be an additional check and only in case the `call_uuid` is not set.


bhepp:
I included a signed CONTRIBUTORS.md with this PR.

danbogos:
Thanks!

PR #362, opened 2016-02-05
Added error details on connections

eloycoto:
Hi, 

Small PR to get an error with details when cgr-loader db type is not valid. 

Regards.


danbogos:
Thanks!

PR #366, opened 2016-02-19
handle extra_fields variables with int/float value

bhepp:
Some channel and CDR variables have int or float values.


danbogos:
Thanks, sorry for delay but just not enough hours in a day lately ;).

DanB

PR #430, opened 2016-04-15
Added mi_fifo module in tutorial

Joonake:
Added mi_fifo module in the kamailio.cfg of the kamevapi tutorial in order to make kamctl to work.


Joonake:
Oops, apparently this caused some problems with Kamailio 4.3, so no need to reload.

PR #446, opened 2016-05-09
Adding error case to have correct value in template without prefix

rbarrabe:
Adding error case to have correct value in template without prefix


danbogos:
Thanks!

PR #448, opened 2016-05-16
OpenSIPS ASR fix, LoadTariffPlanFromStorDb crash fix, missing build dependency in glide.yaml

J0hnSteel:
Hello DanB, just a few fixes.


danbogos:
Hi @J0hnSteel 

Thanks for your patches. Unfortunately cannot merge them due to tests failing. For the sake of exercise, can u maybe fix the tests also? Command `go test` in the sessionmanager folder should guide you on that.

DanB

PS: Good news is that in the future OpenSIPS guys will take the burden of integration since they will develop their own cgrates.so module within OpenSIPS (similar to rtproxy one).

danbogos:
Thanks!

DanB

PR #466, opened 2016-06-07
Save the conf of cdrstat

rbarrabe:
Hi Dan,
I added a feature to save the conf of cdrstat when we save the stat queue otherwise the state of triggers are not saved.


danbogos:
Hi @rbarrabe 

Thanks for this. @rif  will review it and will get back to you.

DanB

PR #481, opened 2016-07-04
empty_balance_context fix - missing dialplan type.

marcinkowalczyk:
There is an issue with empty_balance_context. On FS 1.6.7 transfer fails with No route to destination error.
When calling uuid_transfer api type of dialplan is missing so transfer cannot be executed.  Proper syntax for uuid transfer between contexts:

freeswitch@freeswitch> uuid_transfer
-USAGE: <uuid> [-bleg|-both] <dest-exten> [<dialplan>] [<context>]

This pull request adds XML in request.


danbogos:
Thanks!

DanB

PR #498, opened 2016-07-31
small uuid fix

J0hnSteel:
Hello DanB I noticed that every time I use ApierV1.SetBalance with BalanceUUID , that value was ignored and utils.GenUUID() was used to create a new one.

### DIFF
--- a/engine/balance_filter.go
+++ b/engine/balance_filter.go
@@ -293,6 +293,9 @@ func (bf *BalanceFilter) ModifyBalance(b *Balance) {
 	if b == nil {
 		return
 	}
+        if bf.Uuid != nil {
+		b.Uuid = *bf.Uuid
+	}
 	if bf.ID != nil {
 		b.ID = *bf.ID
 	}


rif:
Hi @J0hnSteel sorry it took so long to look into tour pull request.

The Uuid field was intentionally ignored because the user can set the ID to identify a balance. Uuid is a internally generated identifier that is guaranteed to be unique and consistent.

Do you have a particular use case where you need to set the Uuid?

J0hnSteel:
Hello @rif my particular case is when I make some changes to CGR or just test workflow I have a test account where I expect the test balance to be under common uuid that I can reference from somewhere else (and I destroy and recreate this account often). The other use case that I have in mind is if you save that info in a DB or somewhere else for ease of management or for backup,migration.

rif:
Can't you refer the balance by ID instead of UUID?

J0hnSteel:
Hello rif I re-evaluate my test script and it was OK using balance by ID, so this pull request can be disregarded.

PR #511, opened 2016-08-15
CDRC: Fix issue with disabled cdrc clients

eloycoto:
Hey @danbogos 

This fix the problem if the ID is not *default and default is disabled didn't start and raise an error on process the CDR. 

On the other hand, added a log in case of no enable CDRC clients. 

Regards


danbogos:
Thanks @eloycoto ! 
The issue is not really there (since you still cannot configure *default instance or have config inheritance over multiple files) but it was a good catch (copy paste mistake).
Will push during day another fix which should properly handle CDRC configs inheritance.

DanB

eloycoto:
@danbogos 

I found that problem during today, and I fixed :-).

Nice to know that you're working on the cdrc config, maybe I'll push another patch today (just debugging)

Regards.

PR #512, opened 2016-08-15
Storage: Accounts fields length consistency

eloycoto:
Hey, 

Today I was running some problems with the account names. In account actions the varchar length is 64, but in another places are 24. So I changed all to 64, like 24 is too small for account name. The same for subject. 

In the other hand, in cdr stats, I added 255, that stat is a rsr field. This value should be enough based on the regexp that we have. 

Regards.


danbogos:
Hi @eloycoto 

Please change the MySQL ones also for consistency.

Ta,
DanB

eloycoto:
Sorry! done ;-)

danbogos:
Thank you!
DanB

PR #551, opened 2016-09-24
Redis: Added method to reconnect on redis if any IOErr reply. Fix #543

### LINKED ISSUE #543: Issues on redis timeouts/restart
Hey! 

Related to the issues that we have with Redis. So the main problem that we have is when connection did not reconnect when Redis restarts. 

These are the steps to reproduce the issue: 

1- Start Redis
2- Start cgr-engine process
3- Restart Redis
4- Make any API call. 

When this happens, the result is the following: 

```
{"id":"9bd4506a-7c16-11e6-9806-005056a942b8","result":null,"error":"SERVER_ERROR: EOF"}
```
(The example call is ApierV1.LoadTariffPlanFromStorDb)

So, I tried to fix this but is a bit painful. The problem is on Radix.V2 driver.  When Pool.CMD or Pool.Get; Pool.CMD is executed the system didn't check the connection is correct or not. 

Pool.Conn is mainly Redis.Client structure, so from Pool.Conn.Conn can get the TCP status, but all into this are private methods and can't be used at all. 

I tried to make a function that wraps Pool.Cmd as you proposed yesterday, but there is a problem: the error is not different when is a TCP problem or any Redis exception. So if you send a wrong command, you'll never know about that. 

I made today a few workarounds, and I did not find the correct way. For me, the best option is to enable ping on Redis:  you can add, remove, update pools as you want and will work correctly. As you send a ping, you'll know that all the platform works that Redis always return with pong. 

On the other hand, today, meanwhile I was checking the code and I realised that the cgr-engine create two instances of RedisStorage and the system always create two times the maxConns in the system. In our case 200 instead 100. 

So, we do not want to have a fork of this with pings. So if you can advice me about is the solution that you want I'll try to put that in place. 

Regards.

eloycoto:
Hey @danbogos 

As we spoke in the issue. This fix the issue with Redis. 

Regards.

### DIFF
--- a/engine/storage_redis.go
+++ b/engine/storage_redis.go
@@ -22,14 +22,13 @@ import (
 	"compress/zlib"
 	"errors"
 	"fmt"
-	"io/ioutil"
-	"strings"
-
 	"github.com/cgrates/cgrates/cache2go"
 	"github.com/cgrates/cgrates/config"
 	"github.com/cgrates/cgrates/utils"
 	"github.com/mediocregopher/radix.v2/pool"
 	"github.com/mediocregopher/radix.v2/redis"
+	"io/ioutil"
+	"strings"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


danbogos:
Thanks, will try to find the time during next days to review it.

DanB

danbogos:
hey @eloycoto 

Sorry for delay. I have decided to merge and refactor it a bit on my side.

Thanks for it!

DanB

PR #554, opened 2016-09-28
Adding ExtraFields when we fork cdr

rbarrabe:
We must add the extrafields in forked cdr because this is used to create the final cdr


danbogos:
Thanks!

DanB

PR #558, opened 2016-10-05
/var/log/cgrates directory is no longer created and used

razvancrainea:
Since the directory is no longer created, the postinstall script fails
and the package cannot be installed.

A temporary fix is to manually create the directory, but it is no longer
used anyway in the config file.


danbogos:
Thanks!

DanB

PR #559, opened 2016-10-05
add support for mongo 2.4

razvancrainea:
createUser is only available in 2.6, so for older mongo versions we need
to use deprecated addUser


danbogos:
As previously discussed we will not support older versions of Mongo so this PR is not longer needed.

DanB

PR #564, opened 2016-10-16
Adding connect fee in a increment

rbarrabe:


danbogos:
Decided together to cancel it for now.

DanB

PR #573, opened 2016-10-24
Add command line flag to cgr-engine for logging levels

shaneneuerburg:
This adds a new -log_level flag to cgr-engine, defaulting to 5 (INFO). The logger has been extended to drop log messages that do not meet or exceed the log level requested.

The default is currently set to INFO to catch all existing log calls that are being used. If this PR is accepted, I'd recommend bumping up existing log calls to higher levels and changing the default to not include INFO if you want to limit logging done.

I see this being a decent middle ground where you could go from a firehose of logging to virtually none if desired.


danbogos:
Thanks, just be conservative about adding too much logging calls (so we don't do that many ifs around the code).

DanB

PR #591, opened 2016-11-10
Add basic authentication to HTTP API

shaneneuerburg:
This adds support for basic authentication for the HTTP API with new configuration values to control if basic auth is used, the basic auth realm URL, and the location to a basic .htpasswd file.

github.com/abbot/go-http-auth is added as a dependency as it efficiently implements basic auth, as well as .htpasswd file monitoring.


danbogos:
Hi @shaneneuerburg 

As discussed, we should go for some changes on this one so will close it for a new PR in the future.

DanB

PR #595, opened 2016-11-18
Cloning action for each account

rbarrabe:
Cloning action for each account to be sure when we execute an actionplan that the action cdrlog log the previous correctly


danbogos:
As discussed, please create a new PR using the cache.GetCloned method.

Thanks,
DanB

PR #659, opened 2017-04-05
Update opensips

jeffreywilkie:
referenced opensips version

### DIFF
--- a/data/tutorials/osips_async/opensips/etc/default/opensips
+++ b/data/tutorials/osips_async/opensips/etc/default/opensips
@@ -1,5 +1,5 @@
 #
-# OpenSIPS startup options
+# OpenSIPS 2.2.x startup options 
 #
 
 # Set to yes to enable opensips, once configured properly.


danbogos:
Hi @jeffreywilkie 

Thanks for this! Please sign contributors so I can merge your contribution.

Ta,
DanB

danbogos:
One more thing, please have a look here:
http://cgrates.readthedocs.io/en/latest/tut_opensips_installs.html

It specifically installs 2.2, so no need of adding any additional info in this case.

DanB

jeffreywilkie:
Ok thanks. I was just searching through documents for key words, then reading the documents to see if they needed updating. I would assume any out of date docs would be archived?  I'm attempting to update information that has snagged me while learning.

cgrates:
Sure, np.
Please note that also there is no need of any mongo/postgres info inside
the tutorial since they are built for Redis/MySQL. We would need to provide
another tutorial (different configurations) for other combination of
databases (the one which you use actually).

DanB

Issue #672, opened 2017-05-22
sm_freeswitch is not sending calls to empty_balance_context anymore


marcinkowalczyk:
Hi,

I've upgraded RC8 from last year (Apr 11 2016) to latest one  (22 May 2017) and seems empty_balance_context nor empty_balance_ann_file does not work anymore. In Apr version if there is balance empty SM_Freeswitch transfers call to context defined in empty_balance_context

sm configured following way:

```
 "sm_freeswitch": {
        "enabled": true,    
        "rals_conns": [
                {"address": "*internal"}        
        "cdrs_conns": [
                {"address": "*internal"}       
        ],
        "create_cdr": true,                            
        "extra_fields": ["sip_h_X-CallID","variable_cgr_callerid"],.
        "debit_interval": "10s",
        "min_call_duration": "0s",
        "max_call_duration": "3h",
        "min_dur_low_balance": "0s",
        "low_balance_ann_file": "",
        "empty_balance_context": "empty_balance_context",
        "empty_balance_ann_file": "/usr/share/freeswitch/sounds/speech_20170519112509659",
        "subscribe_park": true,
        "channel_sync_interval": "5m",
        "max_wait_connection": "2s",
        "event_socket_conns":[
                {"address": "10.23.138.24:8021", "password": "ClueCon", "reconnects": 5}
        ],
 },
```

dump from ESL connection

```
#
T 10.23.138.24:8021 -> 10.23.142.98:38958 [AP]
Content-Length: 5894
Content-Type: text/event-plain


#
T 10.23.138.24:8021 -> 10.23.142.98:38958 [AP]
Event-Name: CHANNEL_PARK
Core-UUID: 3b010f11-5df0-4a06-8436-9f464cec50ab
FreeSWITCH-Hostname: aws-test-freeswitch-mock
FreeSWITCH-Switchname: freeswitch_farm_01
FreeSWITCH-IPv4: 10.23.138.24
FreeSWITCH-IPv6: %3A%3A1
Event-Date-Local: 2017-05-22%2015%3A52%3A37
Event-Date-GMT: Mon,%2022%20May%202017%2013%3A52%3A37%20GMT
Event-Date-Timestamp: 1495461157829770
Event-Calling-File: switch_ivr.c
Event-Calling-Function: switch_ivr_park
Event-Calling-Line-Number: 950
Event-Sequence: 24429
Channel-State: CS_EXECUTE
Channel-Call-State: RINGING
Channel-State-Number: 4
[...]

PR #675, opened 2017-05-25
This PR closes issue #672 - sm_freeswitch is not sending calls to empty_balance_context anymore 

marcinkowalczyk:
If there is an empty_balance_context configured sm_freeswitch should send this call to this context not disconnect active call via sched_hangup.


danbogos:
Thanks!

DanB

PR #679, opened 2017-05-31
proposition for new timing tags *minute and *hour to be used in action

J0hnSteel:
plans on recurring charges.
NOTE: most of the time minute interval executes every second minute but logs from the
scheduler report scheduled execution every minute.


danbogos:
@J0hnSteel , thanks for this! 
I have done small customization (like *every_minute and *hourly) but kept most of the logic you have implemented, so good job!

PR #709, opened 2017-08-29
StatsConfig and ResourceConfig

TeoV:
Add methods for StatsConfig: get, set and remove
Add methods for ResourceConfig : get,set and remove
Add Test for methods and add test for update


danbogos:
Thanks!

PR #728, opened 2017-09-10
Fix TPUsers and added test for them 

TeoV:
Set on conf/samples/tutmysql historys false because got an error : 
<HistoryServer> Could not start, error: open /var/lib/cgrates/history/destinations.json: no such file or directory

and engine don't start with this configuration


danbogos:
Thanks!

PR #738, opened 2017-09-15
Add new field("ID") in models TPResource  and replace primary key with PK and set tenant after ID in csv

TeoV:


danbogos:
Thanks!

PR #773, opened 2017-10-10
RequestFilter -> Filter and delete FilterProfile

TeoV:
+ Add get/set/rem for thresholdProfile in apier/v1/thresholds.go


danbogos:
Thanks!

PR #859, opened 2017-12-07
 added tests for migrator

Edwardro22:


migrator/alias.go
```diff
@@ -50,36 +51,43 @@ func (m *Migrator) migrateCurrentAlias() (err error) {
...
 
-func (m *Migrator) migrateCurrentReverseAlias() (err error) {
-	var ids []string
-	ids, err = m.dmIN.DataDB().GetKeysForPrefix(utils.REVERSE_ALIASES_PREFIX)
-	if err != nil {
-		return err
-	}
-	for _, id := range ids {
-		idg := strings.TrimPrefix(id, utils.REVERSE_ALIASES_PREFIX)
+// func (m *Migrator) migrateCurrentReverseAlias() (err error) {
```
danbogos: If we don't longer need this, please remove from sources instead of  leaving it commented.

migrator/migrator_it_test.go
```diff
@@ -1000,101 +1110,1843 @@ func testMigratorStats(t *testing.T) {
...
+			t.Errorf("Expecting: %+v, received: %+v", th, result)
+		}
+		thpr, err := mig.dmOut.GetThresholdProfile(thp.Tenant, thp.ID, true, utils.NonTransactional)
+		if err != nil {
+			t.Error("Error when getting thresholdProfile ", err.Error())
+		}
+		if !reflect.DeepEqual(thp.ID, thpr.ID) {
+			t.Errorf("Expecting: %+v, received: %+v", thp.ID, thpr.ID)
+		}
+	case dbtype == Move:
```
danbogos: dbtype == Move needs review since it does not look too logic here. It should be more like migrator action = move

utils/consts.go
```diff
@@ -562,14 +562,14 @@ const (
 	MetaSubscribers                 = "*Subscribers"
 	MetaDerivedChargersV            = "*DerivedChargers"
 	MetaTpRatingPlans               = "*TpRatingPlans"
-	MetaTpLcrs                      = "*TpLcrs"
+	MetaTpLCRRules                  = "*TpLcrs"
```
danbogos: all meta instructions need to be written as commands (ie here we should have *tp_lcrs

danbogos:
Thanks!

PR #904, opened 2018-01-14
Added tests

Edwardro22:


danbogos:
Thanks!

utils/cgrevent_test.go
```diff
@@ -162,3 +162,31 @@ func TestCGREventFieldAsFloat64(t *testing.T) {
 		t.Errorf("Expecting: %+v, received: %+v", 0, answ)
 	}
 }
+func TestCGREventClone(t *testing.T) {
```
danbogos: Please add testing of result being a real clone and not pointing towards same original object in any of the fields.

PR #906, opened 2018-01-16
Add time.Sleep after each update for mongo

TeoV:


engine/onstor_it_test.go
```diff
@@ -759,6 +759,9 @@ func testOnStorITRatingPlan(t *testing.T) {
 			StartTime: "00:00:00",
 		},
 	}
+	if onStor.DataDB().GetStorageType() == utils.MONGO {
+		time.Sleep(5 * time.Millisecond)
```
danbogos: Please use a global variable for this, so we can tweak it generally.

danbogos:
Thanks!

PR #934, opened 2018-01-31
SM_Cost -> SessionsCost and update FSEvent and KamEvent

TeoV:


agents/fsevent.go
```diff
@@ -399,11 +405,17 @@ func (fsev FSEvent) V1AuthorizeArgs() (args *sessions.V1AuthorizeArgs) {
 
 // V1InitSessionArgs returns the arguments used in SessionSv1.InitSession
 func (fsev FSEvent) V1InitSessionArgs() (args *sessions.V1InitSessionArgs) {
```
danbogos: please add timezone here as argument to function, so we can use it directly from where we have config.

danbogos:
Thanks!

PR #938, opened 2018-02-01
Update Kamailio and Freeswitch agent (SessionSv1ProcessCDR need CGREevent as parameter)

TeoV:


agents/fsagent.go
```diff
@@ -236,7 +236,7 @@ func (sm *FSsessions) onChannelHangupComplete(fsev FSEvent, connId string) {
 	}
 	if sm.cfg.CreateCdr {
 		cdr := fsev.AsCDR(sm.timezone)
```
danbogos: the right fix would be fsev.AsCGREvent. Please add it on todo for the future.

danbogos:
Thanks!

PR #951, opened 2018-02-12
Dynamic usage_ttl for resources

TeoV:


engine/resources_test.go
```diff
@@ -588,7 +589,7 @@ func TestCachedResourcesForEvent(t *testing.T) {
 }
 
 func TestRSmatchingResourcesForEvent(t *testing.T) {
```
danbogos: please add a test to cover all 4 cases of usageTTL (out of API, duration 0 out of API and not 0 out of resource profile and 0 out of resource profile.

danbogos:
Thanks!

PR #1045, opened 2018-03-30
 Add MaxCost for SupplierSv1.GetSuppliers

TeoV:


engine/spls_leastcost.go
```diff
@@ -35,21 +35,36 @@ type LeastCostSorter struct {
 	spS     *SupplierService
 }
 
-func (lcs *LeastCostSorter) SortSuppliers(prflID string,
-	suppls []*Supplier, ev *utils.CGREvent) (sortedSuppls *SortedSuppliers, err error) {
+// LeastCostSorter sorts suppliers based on their cost
+type extraOptions struct {
+	maxCost      *float64
```
danbogos: thinking better, maxCost = 0 does not make much sense, so please make it float64 instead of *float64

danbogos:
Thanks!

PR #1047, opened 2018-03-30
Update MaxCost with *event_cost

TeoV:


engine/suppliers.go
```diff
@@ -302,12 +306,118 @@ func (spS *SupplierService) sortedSuppliersForEvent(args *ArgsGetSuppliers) (sor
...
+	out.ignoreErrors = args.IgnoreErrors
+	return
+}
+
+type optsGetSuppliers struct {
+	ignoreErrors bool
+	maxCost      float64
+}
+
+func (args *optsGetSuppliers) costForEvent(ev *utils.CGREvent, suppls []*Supplier) (val float64, err error) {
```
danbogos: Here I find a bit overloaded to create costForEvent on opts. Maybe we can make a simple function doing it, so we can reuse the code later.

PR #1058, opened 2018-04-20
Changed ping strings to lower case, added sessionSv1 commands to console

Anevo:


console/ping.go
```diff
@@ -48,21 +49,21 @@ func (self *CmdApierPing) Name() string {
 
 func (self *CmdApierPing) RpcMethod() string {
 	switch self.rpcParams.Item {
```
danbogos: Please make the Item to lower and not the suppliers since that is a constant and we don't need to loose processing on run time to make it low.

danbogos:
Thanks!

PR #1064, opened 2018-04-24
Fix Overwrite for ratingprofile_set console command

cruzccl:
This commit fixes rpcParams type so that Overwrite attribute gets
usable through ratingprofile_set command.


danbogos:
Thanks!

PR #1083, opened 2018-05-15
Added encoding tests in migrator

Anevo:


data/conf/samples/tutmysqlmsgpack/cgrates.json
```diff
@@ -0,0 +1,67 @@
+{
```
danbogos: this file is not needed since the default tutmysql works with msgpack

data/conf/samples/tutmongomsgpack/cgrates.json
```diff
@@ -0,0 +1,69 @@
+{
```
danbogos: this file is not needed since tutmongo points out default to msgpack

danbogos:
Thanks!

PR #1087, opened 2018-05-17
Migrator: Added missing .json config options, updated migrator integration tests.

Anevo:


data/conf/samples/tutmongojson/cgrates.json
```diff
@@ -18,7 +18,7 @@
 
 "data_db": {
 	"db_type": "mongo",
-	"db_name": "10",
+	"db_name": "11",
```
danbogos: we can probably add a normal name here, like "cgrates2"

danbogos:
Thanks!

PR #1105, opened 2018-06-13
 Category field as 64 characters in storDb Fixes #254

### LINKED ISSUE #254: Category field as 64 characters in storDb
Since it is used in more and more cases, there are scenarios when it is also bigger than 32.

DanB

TeoV:


danbogos:
Thanks!

PR #1122, opened 2018-06-28
In sessions release resource if was allocated

TeoV:


sessions/sessions.go
```diff
@@ -1694,9 +1694,13 @@ func (smg *SMGeneric) BiRPCv1InitiateSession(clnt rpcclient.RpcClientConnection,
 		if smg.rals == nil {
 			return utils.NewErrNotConnected(utils.RALService)
 		}
-		originID, err := args.CGREvent.FieldAsString(utils.OriginID)
-		if err != nil {
-			return utils.NewErrMandatoryIeMissing(utils.OriginID)
+		var err error
+		originID := ""
```
danbogos: I would keep it named as resourceID since we can understand it's purpose easier. Also in this case, direct assignment could mislead the purpose of the var, I find "var resID string" more like a statement for the purpose of the var - unused yet.

danbogos:
Thanks!

PR #1136, opened 2018-07-09
Attributes process event now accept AttributeIDs and update glide.lock with new version of cgrates/mgo

TeoV:


danbogos:
Thanks!

PR #1157, opened 2018-07-24
AttributeS connection to SupplierS and add  AttributeProfile.Blocker + tests

TeoV:


config/config.go
```diff
@@ -399,12 +398,6 @@ func (self *CGRConfig) checkConfigSanity() error {
 				return errors.New("Alias server not enabled but requested by RALs component.")
 			}
 		}
```
danbogos: please also add sanity check for attributes.suppliers_conns

danbogos:
Thanks!

PR #1158, opened 2018-07-24
engine.ratingplan: check continuity for each destination

### LINKED ISSUE #1153: timings: same destination prefix with different tags makes CGRateS crash
Hi CGRateS team,

We are having some issues when using timings to assign different rates to calls depending on the time they are established. We are not sure if we are using CGRateS incorrectly or if this may be a bug.

We have this minimal configuration running in CGRateS (version e3083471):

- Destinations.csv:
```
DST_34,+34
DST_34_DUP,+34
```
Same prefix, different tag, but fulfilling [database unique](https://github.com/cgrates/cgrates/blob/d829b17745e511001a6fae26f1919096bcace2f3/data/storage/mysql/create_tariffplan_tables.sql#L35). The reason behind this prefix duplication was a design decision we made for making CGRateS integration easier. We haven't experienced any problem until now.

- Rates.csv:
```
RT_10CNT,0.2,0.1,60s,60s,0s
RT_20CNT,0.4,0.2,60s,60s,0s
```
2 different rates.

- DestinationRates.csv:
```
DR_34_10CNT,DST_34,RT_10CNT,*up,4,0,
DR_34_DUP_20CNT,DST_34_DUP,RT_20CNT,*up,4,0,
```
2 different destination-rates to relate previous items.

- Timings.csv:
```
AFTERNOON,*any,*any,*any,*any,12:00:00
```
1 timing for calls after 12:00.

- RatingPlans.csv:
```
RP_GENERIC,DR_34_10CNT,*any,10
RP_GENERIC,DR_34_DUP_20CNT,AFTERNOON,20
```
One rating plan to achieve this:
- Calls to +34 prefix before 12:00, get rated by RT_10CNT (through DR_34_10CNT).
- Calls to +34 prefix after 12:00, get rated by RT_20CNT (through DR_34_DUP_20CNT).

- RatingProfiles.csv:
```
*out,cgrates.org,call,*any,2014-01-14T00:00:00Z,RP_GENERIC,,
```
One rating profile using previous rating plan.

- AccounActions.csv:
```
cgrates.org,1001,,,,
```
One account/subject in default tenant.

With this config, invoking this command causes segfault:
```
cgr-console 'cost Tenant="cgrates.org" Subject="1001" Category="call" AnswerTime="2015-03-10T08:00:00Z" Destination="+34676000000" Usage="60s"'
```

The reason behind this segfault seems to be in [GetRatingPlansForPrefix](https://github.com/cgrates/cgrates/blob/2f7b77f3a14d2f3b4c4e3a06f9f9a4abf4301e0f/engine/ratingprofile.go#L154):
- It detects two different destination tags, so it does not consider them 2 complementary destination-rates and evaluates them separately ordered by weight (the higher, first).

- That causes DR_34_DUP_20CNT been selected first, that is not capable to rate a call made at 8:00 and causes the crash.

Using two different destination tags with the same destination prefix is not supported and [database unique](https://github.com/cgrates/cgrates/blob/d829b17745e511001a6fae26f1919096bcace2f3/data/storage/mysql/create_tariffplan_tables.sql#L35) should be updated to avoid this or timings logic needs a patch to support them?

Sorry for the length of the issue and thank you very much for your time and your awesome project :)

cruzccl:
This commit aims to modify rating plan continuity validation so that
each destination involved in a rating plan is evaluated independently.

This way, if a rating plan is loaded with -validate flag, it will warn
if any of the destinations is not usable all weekdays, avoiding possible
runtime errors.

Fixes #1153

### DIFF
--- a/engine/ratingplan.go
+++ b/engine/ratingplan.go
@@ -96,38 +96,55 @@ func (rp *RatingPlan) Equal(o *RatingPlan) bool {
 	return rp.Id == o.Id
 }
 
-// IsValid determines if the rating plan covers a continous period of time
+// IsValid determines if the rating plan covers a continous period of time for all destinations
 func (rp *RatingPlan) isContinous() bool {
+	if len(rp.DestinationRates) == 0 {
+		return false
+	}
+
 	weekdays := make([]int, 7)
-	for _, tm := range rp.Timings {
-		// if it is a blank timing than it will match all
-		if tm.IsBlank() {
-			return true
+	for _, dr := range rp.DestinationRates {
+		var tmgs []*RITiming
+		for _, tmg := range dr {
+			tmgs = append(tmgs, rp.Timings[tmg.Timing])
 		}
-		// skip the special timings (for specific dates)
-		if len(tm.Years) != 0 || len(tm.Months) != 0 || len(tm.MonthDays) != 0 {
-			continue
-		}
-		// if the startime is not midnight than is an extra time
-		if tm.StartTime != "00:00:00" {
-			continue
-		}
-		//check if all weekdays are covered
-		for _, wd := range tm.WeekDays {
-			weekdays[wd] = 1
-		}
-		allWeekdaysCovered := true
-		for _, wd := range weekdays {
-			if wd != 1 {
-				allWeekdaysCovered = false
+
+		continuous := false
+		for _, tm := range tmgs {
+			// if it is a blank timing than it will match all
+			if tm.IsBlank() {
+				continuous = true
+				break
+			}
+			// skip the special timings (for specific dates)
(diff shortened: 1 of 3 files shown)


danbogos:
Hey Carlos,

Unfortunately I cannot modify this part since it might break the way we calculate increments. For the moment if you define a destination, you have the responsibility to make sure that it contains a cost for all time intervals since we assume that.
This could change in the future but since existing rating is in production since years and stable, we don't want to influence that.
A rating redesign is planned for the future but it will not happen in the next half an year.
The only solution in your case I see is to continue on your side with patched version or properly design your rating plans to follow CGRateS philosophy.

DanB

PR #1160, opened 2018-07-25
Remove supplier connection from AttributeS and update check config sanity

TeoV:


engine/libsuppliers.go
```diff
@@ -79,6 +79,17 @@ func (sSpls *SortedSuppliers) SortCost() {
 	})
 }
 
+// SortCost is part of sort interface,
+// sort based on Cost with fallback on Weight
+func (sSpls *SortedSuppliers) SortHighestCost() {
+	sort.Slice(sSpls.SortedSuppliers, func(i, j int) bool {
```
danbogos: this line needs recheck, not sure if we cannot integrate it in the second comparison. Maybe add a test on ==?

danbogos:
Thanks!

PR #1163, opened 2018-07-27
Add QOS supplier strategy + tests

TeoV:


engine/libsuppliers.go
```diff
@@ -92,10 +93,24 @@ func (sSpls *SortedSuppliers) SortHighestCost() {
...
+func (sSpls *SortedSuppliers) SortQOS(params []string) {
 	sort.Slice(sSpls.SortedSuppliers, func(i, j int) bool {
-		//to be added
-		return true
+		for _, param := range params {
+			// skip to next param
+			if sSpls.SortedSuppliers[i].worstStats[param] == sSpls.SortedSuppliers[j].worstStats[param] {
+				continue
+			}
+			if sSpls.SortedSuppliers[i].worstStats[param] == -1 {
```
danbogos: here could kick in also pdd?

engine/spls_qos.go
```diff
@@ -52,23 +53,47 @@ func (lcs *QOSSupplierSorter) SortSuppliers(prflID string, suppls []*Supplier,
...
+				if metric == strings.Split(keyWithID, utils.InInFieldSep)[0] {
+					if val, hasKey := srtData[metric]; !hasKey ||
+						(metric == utils.MetaPDD && val < value) || //worst values
+						(metric != utils.MetaPDD && val > value) {
+						srtData[metric] = value
+						hasMetric = true
+					}
+				}
+			}
+			if hasMetric == false { //if not have populate with default value
```
danbogos: here you can use short value, if !hasMetric {

engine/spls_qos.go
```diff
@@ -52,23 +53,47 @@ func (lcs *QOSSupplierSorter) SortSuppliers(prflID string, suppls []*Supplier,
...
+				srtData[metric] = -1
+				if metric == utils.MetaPDD {
```
danbogos: better do it with a case here since it could be we will have more in the future.

danbogos:
Thanks!

PR #1164, opened 2018-07-31
 Update Suppliers stategy

TeoV:


apier/v1/suppliers_it_test.go
```diff
@@ -673,6 +674,103 @@ func testV1SplSGetQOSSuppliers3(t *testing.T) {
...
+	}
+	eSpls := engine.SortedSuppliers{
+		ProfileID: "SPL_QOS_FILTRED",
+		Sorting:   utils.MetaQOS,
+		SortedSuppliers: []*engine.SortedSupplier{
+			&engine.SortedSupplier{
+				SupplierID: "supplier1",
+				SortingData: map[string]interface{}{
+					"*acd:Stat_1":   11.0,
+					"*acd:Stat_1_1": 11.0,
```
danbogos: would be interesting to see a different *acd here so we can see them properly using worst stat

engine/suppliers.go
```diff
@@ -284,6 +285,100 @@ func (spS *SupplierService) resourceUsage(resIDs []string) (tUsage float64, err
 	return
 }
 
+func (spS *SupplierService) populateSortingData(ev *utils.CGREvent, spl *Supplier, extraOpts *optsGetSuppliers) (srtSpl *SortedSupplier, pass bool, err error) {
+	globalStats := map[string]float64{ //used for QOS strategy
+		utils.Weight: spl.Weight,
```
danbogos: the global stats here are to be used just for stats, the rest of sorting data is taken out of .SortingData

danbogos:
Thanks!

PR #1183, opened 2018-08-24
Move DataProvider,NavigableMap,RsrParser into config package

TeoV:


config/libconfig_json.go
```diff
@@ -574,9 +574,24 @@ type MigratorCfgJson struct {
 }
 
 type FcTemplateJsonCfg struct {
-	Id       *string
-	Type     *string
-	Field_id *string
-	Filters  *[]string
-	Value    *string
+	Id                   *string
```
danbogos: please use ID instead of Id since the govet will complain about it's format.

danbogos:
Thanks!

PR #1196, opened 2018-09-11
Fixes #803

### LINKED ISSUE #803: export *cdrlog type CDR
Hello Dear CGRateS Team!

We try to auto-export CDRs that were produced by execution of action having *cdrlog action type.
No matter which filter we use, they are not exported.
CDRs that are produced by ProcessCDR API are exported OK.
The *cdrlog does not go through online exports (only 
query-able CDRs since the CDR is dumped directly into StorDB). 
Please correct, so it will be possible to auto-export *cdrlog CDRs at action execution.

Thanks,
Michael

TeoV:


cmd/cgr-engine/cgr-engine.go
```diff
@@ -1087,6 +1088,20 @@ func initLogger(cfg *config.CGRConfig) error {
 	return nil
 }
 
+func createCDRConnection(internalCDRSChan chan rpcclient.RpcClientConnection, exitChan chan bool) {
```
danbogos: please rename to schedCDRsConns

config/config.go
```diff
@@ -722,6 +722,14 @@ func (self *CGRConfig) checkConfigSanity() error {
 				utils.DispatcherS, self.dispatcherSCfg.DispatchingStrategy)
 		}
 	}
+	// Scheduler
+	if self.schedulerCfg != nil && self.schedulerCfg.Enabled {
```
danbogos: please remove the need of schedulerCfg.Enabled here since we could be setting up schedCDRsConns even if scheduler is disabled

engine/action.go
```diff
@@ -279,6 +279,7 @@ func parseTemplateValue(rsrFlds utils.RSRFields, acnt *Account, action *Action)
 }
 
 func cdrLogAction(acc *Account, sq *CDRStatsQueueTriggered, a *Action, acs Actions) (err error) {
+
```
danbogos: please return error here as discussed if schedConns is nil

engine/action.go
```diff
@@ -335,13 +339,12 @@ func cdrLogAction(acc *Account, sq *CDRStatsQueueTriggered, a *Action, acs Actio
 			}
 		}
 		cdrs = append(cdrs, cdr)
-		if cdrStorage == nil { // Only save if the cdrStorage is defined
+		if (schedCdrsConns == nil) || (schedCdrsConns != nil && reflect.ValueOf(schedCdrsConns).IsNil()) {
```
danbogos: this is not needed in loop

engine/actions_it_test.go
```diff
@@ -75,8 +75,46 @@ func TestActionsitRpcConn(t *testing.T) {
 	}
 }
 
-func TestActionsitSetCdrlogDebit(t *testing.T) {
+func TestActionsitAddChargers(t *testing.T) {
```
danbogos: as discussed, there is no need of chargers here, please remove this test

danbogos:
Thanks!

PR #1198, opened 2018-09-12
Add dataProvider for cdrLog Action

TeoV:


config/config.go
```diff
@@ -722,14 +722,6 @@ func (self *CGRConfig) checkConfigSanity() error {
 				utils.DispatcherS, self.dispatcherSCfg.DispatchingStrategy)
 		}
 	}
```
danbogos: please keep the check here, just only  "for _, connCfg := range self.schedulerCfg.CDRsConns", without self.schedulerCfg.Enabled

engine/action.go
```diff
@@ -297,10 +161,7 @@ func cdrLogAction(acc *Account, sq *CDRStatsQueueTriggered, a *Action, acs Actio
 			return
```
danbogos: please make sure you are checking for nil schedConns

engine/actions_test.go
```diff
@@ -1468,6 +1467,7 @@ func TestTopupActionLoaded(t *testing.T) {
 	}
 }
 
+/* Disabled tests with cdrLogAction because it need a rpc connection
```
danbogos: please remove completely the commented test, so we don't have useless code there

danbogos:
Thanks!

PR #1201, opened 2018-09-12
 Add check for schedCDRSConns

TeoV:


config/config.go
```diff
@@ -722,6 +722,14 @@ func (self *CGRConfig) checkConfigSanity() error {
 				utils.DispatcherS, self.dispatcherSCfg.DispatchingStrategy)
 		}
 	}
+	// Scheduler check connection with CDR Server
+	if len(self.schedulerCfg.CDRsConns) != 0 {
```
danbogos: should be enough the for here since it will not go into empty/nil conns.

danbogos:
Thanks!

PR #1216, opened 2018-09-20
Modified tests for Session NewV1 and partially added test for mapevent.go

Trial97:


engine/mapevent_test.go
```diff
@@ -0,0 +1,116 @@
...
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>
+*/
+
+package engine
+
+import (
+	"reflect"
+	"testing"
+	//"time"
```
danbogos: Please remove the imports if we don't need them. Leaving them commented will just create garbage.

danbogos:
Thanks!

PR #1218, opened 2018-09-20
Update travis with go 1.11, return error instead of panic ( RSRParser…

TeoV:
…s) and active connection from RALs to StatS

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -33,15 +33,15 @@ In a nutshell all the dependencies are installed and used from a folder named *v
 
 First we have to setup the GO Lang to our OS. Feel free to download 
 the latest GO binary release from https://golang.org/dl/
-In this Tutorial we are going to install Go 1.10.3.
+In this Tutorial we are going to install Go 1.11.0.
 
 ::
 
-   cd $HOME && mkdir -p src/go opt && cd src
-   wget https://dl.google.com/go/go1.10.3.linux-amd64.tar.gz
-   tar -xvf go1.10.3.linux-amd64.tar.gz -C $HOME/opt
-   export GOPATH=$HOME/src/go
-   export PATH=$PATH:$HOME/opt/go/bin:$GOPATH/bin
+   rm -rf /usr/local/go
+   cd /tmp
+   wget https://dl.google.com/go/go1.11.linux-amd64.tar.gz
+   sudo tar -xvf go1.11.0.linux-amd64.tar.gz -C /usr/local/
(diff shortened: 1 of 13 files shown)


docs/installation.rst
```diff
@@ -33,15 +33,15 @@ In a nutshell all the dependencies are installed and used from a folder named *v
 
 First we have to setup the GO Lang to our OS. Feel free to download 
 the latest GO binary release from https://golang.org/dl/
-In this Tutorial we are going to install Go 1.10.3.
+In this Tutorial we are going to install Go 1.11.0.
```
danbogos: Should be here also 1.11 instead of .0

engine/balances.go
```diff
@@ -696,31 +696,35 @@ func (b *Balance) AsBalanceSummary(typ string) *BalanceSummary {
...
+			utils.Account:     acntTnt.ID,
+			utils.BalanceID:   b.ID,
+			utils.Units:       b.Value}}
 	if !b.ExpirationDate.IsZero() {
-		thEv.Event[utils.ExpiryTime] = b.ExpirationDate.Format(time.RFC3339)
+		cgrEv.Event[utils.ExpiryTime] = b.ExpirationDate.Format(time.RFC3339)
+	}
+	if statS != nil {
+		var reply []string
+		go statS.Call(utils.StatSv1ProcessEvent, &StatsArgsProcessEvent{CGREvent: cgrEv}, &reply)
```
danbogos: here should be better to use anonymous function and define reply inside. Also I see you are not treating the error (log it to syslog)

engine/balances.go
```diff
@@ -696,31 +696,35 @@ func (b *Balance) AsBalanceSummary(typ string) *BalanceSummary {
...
 	}
 	var tIDs []string
```
danbogos: tIDs should be defined after if since it is not logical to keep it out

sessions/sessions.go
```diff
@@ -280,7 +280,14 @@ func (smg *SMGeneric) ttlTerminate(s *SMGSession, tmtr *smgSessionTerminator) {
 	}
 	cdr.Usage = s.TotalUsage
 	var reply string
-	smg.cdrsrv.Call("CdrsV1.ProcessCDR", cdr, &reply)
+	cgrEv := &utils.CGREvent{
+		Tenant: s.EventStart.GetStringIgnoreErrors(utils.Tenant),
```
danbogos: Here you should use the s.Tenant directly

danbogos:
Thanks!

PR #1219, opened 2018-09-21
Update account and balance Publish method

TeoV:


engine/account.go
```diff
@@ -1085,12 +1085,19 @@ func (acnt *Account) Publish() {
 			utils.Account:       acntTnt.ID,
 			utils.AllowNegative: acnt.AllowNegative,
 			utils.Disabled:      acnt.Disabled}}
-	var tIDs []string
 	if statS != nil {
 		var reply []string
```
danbogos: please keep together with the API call, inside the anonymous function

engine/balances.go
```diff
@@ -715,10 +715,17 @@ func (b *Balance) Publish() {
 	}
 	if statS != nil {
 		var reply []string
```
danbogos: same here, reply should be within the anonymous function

danbogos:
Thanks!

PR #1242, opened 2018-10-03
Add support for xml from CDRC for indexing

TeoV:


cdrc/xml.go
```diff
@@ -235,6 +236,24 @@ func (xP *xmlProvider) FieldAsInterface(fldPath []string) (data interface{}, err
 	}
 	err = nil                                                 // cancel previous err
 	relPath := utils.HierarchyPath(fldPath[len(xP.cdrPath):]) // Need relative path to the xmlElmnt
+	var slctrStr string
+	for i := range relPath {
+		if sIdx := strings.Index(relPath[i], "["); sIdx != -1 {
+			slctrStr = relPath[i][sIdx:]
```
danbogos: please also add the filtering for items in the same group with value.

cdrc/xml.go
```diff
@@ -235,6 +236,24 @@ func (xP *xmlProvider) FieldAsInterface(fldPath []string) (data interface{}, err
...
+			if slctrStr[len(slctrStr)-1:] != "]" {
+				return nil, fmt.Errorf("filter rule <%s> needs to end in ]", slctrStr)
+			}
+			relPath[i] = relPath[i][:sIdx]
+			if slctrStr[1:2] != "@" {
```
danbogos: need to understand why not [:1] ?

PR #1253, opened 2018-10-19
Migrate from glide to dep

ciarant:


ciarant:
No worries Dan. Thanks for the note and apologies for the distraction.

PR #1261, opened 2018-10-23
Protection against panic

chrisgis333:
Instead panic return error if request_fields of request_processor is empty.

### DIFF
--- a/agents/diamagent.go
+++ b/agents/diamagent.go
@@ -206,6 +206,9 @@ func (da *DiameterAgent) processRequest(reqProcessor *config.DARequestProcessor,
 		return
 	}
 	cgrEv := agReq.CGRRequest.AsCGREvent(agReq.tenant, utils.NestingSep)
+	if cgrEv == nil {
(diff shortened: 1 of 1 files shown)


danbogos:
hey @chrisgis333 

The reason we left that one unprotected was because it is not supposed to be ever empty. Can you give me an example of how to reproduce the panic? Maybe open a separate issue so we can understand better the problem.

Ta,
DanB

chrisgis333:
It is not a problem.
I run on this simply by typo in configuration.
However typos can happened even in production environment.
And in my opinion error is 100 times better than crash in production.

An empty request_fields can occur when I want to simply resend OK in answer without any processing.

danbogos:
OK, even for that, can u open an issue (with typo in config)?

Thanks,
DanB

danbogos:
hey @chrisgis333 

Since there was no feedback, will close the PR. Please reopen if you manage to reproduce the issue with some test.

Thanks again for your efforts!

DanB

PR #1273, opened 2018-10-31
Adding suport for envVar and group config

Trial97:


config/rjreader.go
```diff
@@ -0,0 +1,342 @@
...
+	"bufio"
+	"io"
+	"os"
+
+	"github.com/cgrates/cgrates/utils"
+)
+
+/*
+ *ToDo:
+ *	-add errors to util 	x
```
danbogos: please remove the ToDo from here since they will stay forever otherwise

danbogos: We can add them in issues

danbogos:
Thanks!

PR #1312, opened 2018-11-20
Add logging in case of incorrect amqp credentials fixes #644

### LINKED ISSUE #644: No logging in case of incorrect amqp credentials
When you're using incorrect username / password details for a AMQP connection there is no error in the logfile.

TeoV:


danbogos:
Thanks!

PR #1316, opened 2018-11-21
When attributes is require and isn't found return mandatory ie missin…

### LINKED ISSUE #1315: *attribute as a mandatory substitution
Hello Dan,
Just a reminder for using *attribute as a mandatory substitution.
In case attribute is not found, it is not replied to diameter, so is allowing to pass.

Please fix.
Thanks,
Michael

TeoV:
…g fixes #1315


danbogos:
Thanks!

PR #1319, opened 2018-11-22
Added new filter types

Trial97:


engine/filters.go
```diff
@@ -30,8 +30,11 @@ import (
 )
 
 const (
+	MetaNot            = "*not"
```
danbogos: Here we need multiple filters: *notsuffix, *notstring, *notempty

danbogos:
Thanks!

PR #1326, opened 2018-11-26
Capitalize RATED constant

scottlangendyk:
Capitalize the utils.RATED constant to be in line with the docs for CDR Server https://github.com/cgrates/cgrates/blob/master/docs/cdrserver.rst


danbogos:
hi @scottlangendyk 

thanks for this!

In this case the docs are the one being outdated, the right use of rated type would be "*rated" in the code.

Please feel free to do a PR with docs update.

Thanks,
DanB

PR #1330, opened 2018-12-03
Fix cgrates not recognizing negative UTC offsets in valid RFC3339 datetime format

powerpbx:
We are using debian 9 to reproduce.  Assuming you have a MySQL storDB configured with a row that has tpid="b1"

```
timedatectl set-timezone America/Los_Angeles
systemctl restart cgrates

# cgr-console
cgr> load_tp_from_stordb TPid="b1" FlushDb=false DryRun=true Validate=false Cleanup=false
Error executing command: SERVER_ERROR: cannot parse activation time from 2018-11-24T08:00:00-08:00
exit

timedatectl set-timezone America/New_York
systemctl restart cgrates

# cgr-console
cgr> load_tp_from_stordb TPid="b1" FlushDb=false DryRun=true Validate=false Cleanup=false
Error executing command: SERVER_ERROR: cannot parse activation time from 2018-11-24T08:00:00-05:00
exit

timedatectl set-timezone UTC
systemctl restart cgrates

# cgr-console
cgr> load_tp_from_stordb TPid="b1" FlushDb=false DryRun=true Validate=false Cleanup=false
"OK"
exit

timedatectl set-timezone Europe/Berlin
systemctl restart cgrates

# cgr-console
cgr> load_tp_from_stordb TPid="b1" FlushDb=false DryRun=true Validate=false Cleanup=false
"OK"
```

### DIFF
--- a/utils/coreutils.go
+++ b/utils/coreutils.go
@@ -257,7 +257,7 @@ func ParseDate(date string) (expDate time.Time, err error) {
 			}
 			expDate = expDate.Add(extraDur)
 		}
-	case strings.HasSuffix(date, "Z") || strings.Index(date, "+") != -1: // Allow both Z and +hh:mm format
+	case strings.HasSuffix(date, "Z") || strings.Index(date, "+") != -1 || strings.Index(date, "-") != -1: // Allow both Z and +-hh:mm format
 		expDate, err = time.Parse(time.RFC3339, date)
 	default:
 		unix, err := strconv.ParseInt(date, 10, 64)


powerpbx:
Won't work.  Closing until we work out how to isolate the timezone offset at the end.

powerpbx:
This code will check for negative timezones.

package main

import (
    "fmt"
    "time"
    "reflect"
    "math"
)

func main() {
    date := "2018-12-03T22:23:22-01:00"
    t, err := time.Parse(time.RFC3339, date)
    _, tzOffset:= t.Zone()
    i := float64(tzOffset)
    fmt.Println(tzOffset)
    fmt.Println("Is it negative: ", math.Signbit(i)) 
    if err != nil {
		fmt.Println("error", err)
		return
	}
}
It looks to us like the ParseDate function needs to be rewritten or another function needs to be added which is then incorporated into that function.

PR #1346, opened 2018-12-14
Update messageSetAVPsWithPath from libdiam in case that we need to ad…

TeoV:
…d a group to the last avp (+tests for it)


danbogos:
Thanks!

PR #1378, opened 2019-01-10
fix cc_usage to comply with RFC4006

anthonysomerset:
RFC4006 states that:

>  The CC-Request-Number AVP (AVP Code 415) is of type Unsigned32 and
>    identifies this request within one session.  As Session-Id AVPs are
>    globally unique, the combination of Session-Id and CC-Request-Number
>    AVPs is also globally unique and can be used in matching credit-
>    control messages with confirmations.  An easy way to produce unique
>    numbers is to set the value to 0 for a credit-control request of type
>    INITIAL_REQUEST and EVENT_REQUEST and to set the value to 1 for the
>    first UPDATE_REQUEST, to 2 for the second, and so on until the value
>    for TERMINATION_REQUEST is one more than for the last UPDATE_REQUEST.

If we subtract 2 from the request number in cgrates we are assuming that we are counting from 1 on the initial request, when in fact RFC stats we should set initial request as 0 and increment initial upgrade request from 1 upwards

when subtracting by 2 we lose 1 debit interval of usage in the final usage value

This is a very simple fix which should solve this


TeoV:
Hi Anthony,

Please also update the Request Number in tests : 
Init: https://github.com/cgrates/cgrates/blob/master/agents/diam_it_test.go#L440
Update: https://github.com/cgrates/cgrates/blob/master/agents/diam_it_test.go#L515
Terminate: https://github.com/cgrates/cgrates/blob/master/agents/diam_it_test.go#L590

Thanks,
TeoV

anthonysomerset:
done, thanks for the heads up

danbogos:
Thanks!

PR #1398, opened 2019-02-06
Added support for AWS SQS. closes #1396

### LINKED ISSUE #1396: Request to support AWS SQS as an export end-point for CDRs
It would be really cool if you could support AWS SQS as an export end-point for CDRs.

You recently added support for export to Amazon MQ which was greatly appreciated. However, we found out that Amazon MQ does not currently support any events that we can hook into in order to launch our Lambda function that will dequeue CDRs from it.

SQS however does have support for events which makes it a much nicer fit for us.

Trial97:


danbogos:
Thanks!

PR #1414, opened 2019-02-15
 Removed AliaseS 

Trial97:


apier/v1/suppliers_it_test.go
```diff
@@ -971,6 +972,60 @@ func testV1SplSRemSupplierProfiles(t *testing.T) {
 	}
 }
 
+func testV1SplSSetSupplierProfilesWithoutRatingPlanIDs(t *testing.T) {
```
danbogos: Please move this test into general_tests since in the apier we should keep structure simple

apier/v1/apier.go
```diff
@@ -342,12 +331,6 @@ func (self *ApierV1) LoadTariffPlanFromStorDb(attrs AttrLoadTpFromStorDb, reply
 		}
 	}
 
-	if len(userKeys) != 0 && self.Users != nil {
```
danbogos: I think this one is not really needed since it is part of users but I guess it will do no harm since we also remove users in the near future

danbogos:
Thanks!

PR #1478, opened 2019-03-25
Update Dockerfile for FreeSWITCH 1.8, install from deb and latest cgr-engine syntax.

ghenry:
The sed commands in here are because the current publish deb bundles in the files I'm trying to change, so this will need a further update to remove these.


danbogos:
Thanks!

PR #1483, opened 2019-03-27
Renamed XWrapper to XWithCache

TeoV:


apier/v1/attributes.go
```diff
@@ -60,13 +60,13 @@ func (apierV1 *ApierV1) GetAttributeProfileIDs(tenant string, attrPrfIDs *[]stri
 	return nil
 }
 
-type AttributeWrapper struct {
+type AttributeWithCache struct {
 	*engine.AttributeProfile
```
danbogos: need to check this with ngrep to see if we can set it as pointer

danbogos:
Thanks!

PR #1489, opened 2019-03-29
Fixes #1419 Add new action type *remove_expired

### LINKED ISSUE #1419: Scheduled action to remove expired balances for not active accounts
Hi,

It would be nice to add scheduled action to remove expired balances for not active accounts.

Thanks

TeoV:


dispatchers/utils.go
```diff
@@ -167,6 +167,12 @@ type AttrRemoteUnlockWithApiKey struct {
 	RefID string
 }
 
+type StringkWithApiKey struct {
```
danbogos: Need better naming here. We should analyze it in a call.

Thanks

danbogos:
Thanks!

PR #1536, opened 2019-05-02
Update integration tests

TeoV:
1) Move GetResource from ApierV1 to ResourceSv1 ( because SupplierS and FilterS have a connection to ResourceS) 
2) Corrected in cgr-console Responder method with CallDescriptorWithArgDispatcher.
3)Correctly returned active sessions in case of missing filters


danbogos:
Thanks!

PR #1554, opened 2019-05-15
fixing rsyslog initial install permissions

anthonysomerset:
Found an issue on clean installs where rsyslog can't write to the folder and thus does not log

simple CHOWN during the postinst should sort,


danbogos:
Thanks!

PR #1556, opened 2019-05-17
 Update compose_filter_indexes to rebuild indexes for *any context

TeoV:
1)  In case that we give to compose_filter_indexes context *any for attributes we should rebuild them with the new context, even if the attributes hadn't filter indexed for *any.
1.1) Added test for this case.
2) Added Apier through Dispatcher. For the moment we disabled this until we figured out a better solution in case of GOB server.
2.1)Added test for this.
3) Added benchmark tests( in sessions package) for both server JSON and GOB


danbogos:
Thanks!

PR #1567, opened 2019-05-22
base dictionaries should not sit inside Huawei directory

anthonysomerset:
the base diameter dictionaries shouldn't really sit inside the huawei dictionary directory

in reality we might want to move the vendor specific ones into their own directories as well


danbogos:
Thanks!

PR #1568, opened 2019-05-22
Adding Metaswitch Diameter Dictionary

anthonysomerset:
This enables CGRateS support of the Metaswitch Custom attributes in the Diameter Agent and then able to map it to a field


danbogos:
Thanks!

PR #1580, opened 2019-05-29
default logging folder permissions don't allow writes

anthonysomerset:
So default folder permissions (755) won't allow syslog to write to the folder if owned by root, best solution is to allow group write permissions for the adm group

this allows the ownership to remain as root:adm which are default (in debian and ubuntu) users and groups always available out of the box and for rsyslog to write to the folder accordingly


danbogos:
Thanks!

PR #1583, opened 2019-05-31
 Updated CGRID generation when empty

Trial97:
- Updated AWSPoster.
- Added some unit tests.
- Made OriginHost  to be used uniform for agents.
- Updated CGRID computing from OriginID and OriginHost


danbogos:
Thanks!

PR #1596, opened 2019-06-13
In case of CDRsV1RateCDR allow the CDR to be updated

TeoV:


config/config_defaults.go
```diff
@@ -212,6 +212,7 @@ const CGRATES_CFG_JSON = `
 		"synchronous": false,							// block processing until export has a result
 		"attempts": 1,									// export attempts
 		"field_separator": ",",							// used field separator in some export formats, eg: *file_csv
+		"attributes_context": "",						// context for attributes (empty string means disable attributes process)
```
danbogos: please reformulate the comment to something like: "attributes context - empty disables attributes processing"

danbogos:
Thanks!

PR #1598, opened 2019-06-19
fixing doc refs for cgr-migrate from migrate to exec flag

anthonysomerset:
this is a documentation/feedback change to update the fact that cgr-migrator uses the -exec flag rather than the -migrate flag now


danbogos:
Thanks!

PR #1599, opened 2019-06-19
Prettify duration in cgr-console

Trial97:


engine/mapevent.go
```diff
@@ -78,10 +78,10 @@ func (me MapEvent) GetString(fldName string) (out string, err error) {
 	if !has {
 		return "", utils.ErrNotFound
 	}
-	return utils.IfaceAsString(fldIface)
+	return utils.IfaceAsString(fldIface), nil
 }
 
-func (me MapEvent) GetInt64(fldName string) (out int64, err error) {
+func (me MapEvent) GetTInt64(fldName string) (out int64, err error) {
```
danbogos: Quick question here, qhy GetTInt64?

danbogos:
Thanks!

engine/mapevent.go
Trial97: This function is used to get the connIdx in agents. Because for some reasons the value in map becomes float I added a function which returns int64 even if in event the type is float.

PR #1604, opened 2019-06-24
Added kafka and S3 as cdre

Trial97:


vtzan:
Great news! I will test soon!
thank you guys!

sessions/sessions.go
```diff
@@ -355,9 +355,11 @@ func (sS *SessionS) setSTerminator(s *Session) {
 		case <-endChan:
 			timer.Stop()
 		}
-		s.Lock()
-		s.sTerminator = nil
-		s.Unlock()
+		// this should be done in unregister session to avoid a race condition that can lead to sessions with a fake terminator
```
danbogos: was this removed just for testing? I was thinking we have fixed that part or not?
Need an answer before being able to accept the pr please

Trial97: First it was not removed, only moved in unregisterSession:https://github.com/cgrates/cgrates/pull/1604/files?file-filters%5B%5D=.go&file-filters%5B%5D=.lock&file-filters%5B%5D=.yaml#diff-daebf6f0990b360632bf6eea6800f1dbR778
This was done because when the relocateSessions (https://github.com/cgrates/cgrates/blob/master/sessions/sessions.go#L1190) was called we lock the respective session, unregister it, unlock it and then we registerSession with OriginID changed.  We call unregister wheb the session is locked, we close endChan(https://github.com/cgrates/cgrates/blob/master/sessions/sessions.go#L355) and we switch context to the gorutine. The gorutine will wait(https://github.com/cgrates/cgrates/blob/master/sessions/sessions.go#L358) until session is unlocked to make the Terminator nil. After we unlock the session we call registerSession and here we call setSTerminator. Here we try to get the lock for the session once again(https://github.com/cgrates/cgrates/blob/master/sessions/sessions.go#L286).
Here it's created the race condition because we still need to obtain the lock here and in gorutine. If we obtain the lock in gorutine first everything works as intended. But if we obtain the lock in setSTerminator the terminator for the session is not nill and will not create a new gorutine to monitor for ttl. This can make the session to remain for a undefined time active even if ttl is set in config.
I moved after context switch in unregister session so we do not need to obtain the lock for session once again. 
The integration test are passing for session with this modifications.

danbogos: Unfortunately we cannot remove the session lock in goroutine since that is asking for trouble(no longer control access in random way). Please remove this fix for now, will have to review the session locking once again since you are correct, setSTerminator should not lock the session but have it locked in a layer above.
Thanks!

danbogos:
Thanks!

PR #1605, opened 2019-06-28
For GetCDRs(MySQL) prioritize AnswerTime in query

TeoV:


utils/map.go
```diff
@@ -247,3 +247,32 @@ func MapStringToInt64(in map[string]string) (out map[string]int64, err error) {
...
+		if len(subsystemWithIDs) == 2 {
+			result[subsystemWithIDs[0]] = strings.Split(subsystemWithIDs[1], INFIELD_SEP)
+		} else if len(subsystemWithIDs) > 2 {
+			return nil, ErrUnsupportedFormat
+		}
+	}
+	return result, nil
+}
+
+type MapSubsystemIDs map[string][]string
```
danbogos: I would vote for FlagsWithParams in the form of map[string]interface{} since in the future we could add other params within the flags parsing and by forcing the params to []string we might limit ourselves and be forced to change again. On the FlagsWithParams we should have a method to retrieve the ids, ie: 
`
func (fWp FlagsWithParams) ParamsSlice(subs string) (ps []string) {
    if psIfc, has  := fWp[subs]; has {
        ps, _ = psIfc.([]string)
    return
}
`

sessions/sessions.go
```diff
@@ -2194,20 +2239,25 @@ func (sS *SessionS) BiRPCv1InitiateSessionWithDigest(clnt rpcclient.RpcClientCon
 
 // NewV1UpdateSessionArgs is a constructor for update session arguments
 func NewV1UpdateSessionArgs(attrs, acnts bool,
```
danbogos: Please group here the arguments to look a bit more organized (ie: attributes, attributesIDs, ....)

PR #1608, opened 2019-07-02
Updated kamailio agent

Trial97:


agents/kamevent.go
```diff
@@ -77,7 +77,7 @@ func NewKamEvent(kamEvData []byte, alias, adress string) (KamEvent, error) {
 	if err := json.Unmarshal(kamEvData, &kev); err != nil {
 		return nil, err
 	}
-	kev[utils.CGROriginHost] = utils.FirstNonEmpty(kev[utils.CGROriginHost], alias, adress)
+	kev[utils.OriginHost] = utils.FirstNonEmpty(kev[utils.OriginHost], alias, adress)
```
danbogos: Is there a reason for renaming cgr_originhost to OriginHost? The problem is that we are building the event in Kamailio and there we should have separate naming from standard pseudovariables.

danbogos:
Thanks!

PR #1611, opened 2019-07-09
Add support for *text_plain as reply for HTTPAgent

TeoV:
+ Correctly access in SessionS the AttributeIDs variable


danbogos:
Thanks!

PR #1618, opened 2019-07-15
Use DAEMON_OPTS from default file in systemd service too.

sebastic:
This makes it more in line with the init script with requiring a systemd override for the service.


danbogos:
Thanks!

PR #1625, opened 2019-07-17
Fix logrotate configuration.

sebastic:
logrotate refused to rotate the logfile due to permissions:
```
logrotate[6248]: error: skipping "/var/log/cgrates/CGRateS.log" because parent directory has insecure permissions (It's world writable or writable by group which is not "root") Set "su" directive in config file to tell logrotate which user/group should be used for rotation
```
Ownership of the directory and logfile is root/adm:
```
total 192
drwxrwxr-x 2 root adm    4096 Jul 17 10:24 .
drwxr-xr-x 8 root root   4096 Jul 17 00:00 ..
-rw-r----- 1 root adm  182485 Jul 16 15:52 CGRateS.log
```

### DIFF
--- a/data/conf/logging/deb_logrotate.conf
+++ b/data/conf/logging/deb_logrotate.conf
@@ -6,6 +6,7 @@
         compress
         delaycompress
         sharedscripts
+        su root adm
         postrotate
 			/usr/lib/rsyslog/rsyslog-rotate
         endscript


data/conf/logging/deb_logrotate.conf
```diff
@@ -6,6 +6,7 @@
         compress
         delaycompress
         sharedscripts
```
danbogos: hey Bas,

It appears that adm is not created on all debian versions. Can you add just the root user?

Thanks,
DanB

sebastic: No because the defaults for logrotate are root/root, and rsyslog uses root/adm by default.

Which Debian version are you encountering issues?

The oldest I have available is lenny, and its rsyslog also uses root/adm.

sebastic: An alternative may be to change the rsyslog configuration to use root/root for the cgrates log files.

anthonysomerset:
so we have probably flipped around on this before - the error specifically mentions insecure

root and adm are debian base users always present (in current Debian and Ubuntu distributions at least)

this change makes logical sense for this scenario

danbogos:
Thanks!

PR #1645, opened 2019-07-26
Add DeleteTariffPlanFromFolder API

TeoV:


utils/consts.go
```diff
@@ -583,6 +583,7 @@ const (
 	MetaTenant                   = "*tenant"
 	ResourceUsage                = "ResourceUsage"
 	MetaDuration                 = "*duration"
+	MetaPhoneNumber              = "*phone_number"
```
danbogos: please rename to *libphonenumber so we can be close to google tagging

danbogos:
Thanks!

PR #1646, opened 2019-08-01
 Add new filter type "*accounts" + test

TeoV:


engine/filters.go
```diff
@@ -139,7 +158,18 @@ func (fS *FilterS) Pass(tenant string, filterIDs []string,
...
-			if pass, err = fltr.Pass(ev, fS.statSConns, tenant); err != nil || !pass {
+			// in case we have filters of type *stats,*resources or *accounts we need to send
+			// a rpcclient to get what we need (Metrics,Resource,Accounts)
+			var conn rpcclient.RpcClientConnection
+			switch fltr.Type {
+			case MetaStatS, MetaNotStatS:
+				conn = fS.statSConns
+			case MetaResources, MetaNotResources:
+				conn = fS.resSConns
+			case utils.MetaAccounts:
```
danbogos: please rename the filter to *account so we can keep the name "*accounts" free for future when we will separate into own *accounts subsystem.

Thanks!

danbogos:
Thanks!

PR #1669, opened 2019-08-23
TeoV Add LoaderSv1.Remove API

TeoV:
LoaderSv1.Remove will read from a file( based on a template) line by line and when will find a complete item will delete it from the database and will release the memory. With this will prevent memory overload.


danbogos:
Thanks!

PR #1695, opened 2019-09-23
Add ApierV1.GetRatingPlanIDs + test

TeoV:


config/ralscfg.go
```diff
@@ -33,6 +33,7 @@ type RalsCfg struct {
 	RemoveExpired            bool
 	RALsMaxComputedUsage     map[string]time.Duration
 	RALsBalanceRatingSubject map[string]string
+	RALsMaxIncrements        int
```
danbogos: Since we have now rals in separate config, please rename all it's config variables without the RALs prefix

danbogos:
Thanks!

PR #1706, opened 2019-10-09
Update tests for sessions after simplified locking

TeoV:


sessions/libsessions_test.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 func TestLibSessionSGetSetCGRID(t *testing.T) {
-	sEv := engine.NewSafEvent(map[string]interface{}{
+	sEv := engine.NewMapEvent(map[string]interface{}{
```
danbogos: Please use in all tests engine.MapEvent(map...) instead of New since that one will only apply another if.

danbogos:
Thanks!

PR #1712, opened 2019-10-14
Add delay for sessions integration tests and consider paginator for g…

TeoV:
…et cdr for internal db

### DIFF
--- a/engine/storage_cdrs_it_test.go
+++ b/engine/storage_cdrs_it_test.go
@@ -259,6 +259,15 @@ func testGetCDRs(cfg *config.CGRConfig) error {
 	if err := InitStorDb(cfg); err != nil {
 		return fmt.Errorf("testGetCDRs #1: %v", err)
 	}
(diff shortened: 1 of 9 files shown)


engine/storage_cdrs_it_test.go
```diff
@@ -499,7 +508,7 @@ func testGetCDRs(cfg *config.CGRConfig) error {
 	if CDRs, _, err := cdrStorage.GetCDRs(&utils.CDRsFilter{Paginator: utils.Paginator{Limit: utils.IntPointer(5), Offset: utils.IntPointer(0)}}, false); err != nil {
 		return fmt.Errorf("testGetCDRs #9 err: %v", err)
 	} else if len(CDRs) != 5 {
-		return fmt.Errorf("testGetCDRs #10, unexpected number of CDRs returned: %+v", CDRs)
+		return fmt.Errorf("testGetCDRs #10, unexpected number of CDRs returned: %+v", len(CDRs))
```
danbogos: please remove the prefix testGetCDRs #x from tests and leave only the message since we don't really need it.

PR #1713, opened 2019-10-16
Update tutorial to use GO 1.13 (for mods feature).

sebastic:
The installation instructions were updated to remove the need for `glide` because thats obsolete with the mods feature in Go 1.13, but the instructions still installed Go 1.11.


danbogos:
Thanks!

PR #1725, opened 2019-10-30
Add Remote_conns and Replication_conns options in DataDBConfig

TeoV:
Delete ConfigureDataStorage and add NewDataDBConn which creates a DataDB interface accordingly to config.
Rename ConfigureStorStorage to NewStorDBConn


danbogos:
Thanks!

PR #1734, opened 2019-11-06
Implement Replication Remote mechanism for DataManager for some methods

TeoV:
+ Disable Cache when using internalDB as DataDB
+ When setting Passive session set the session received as parameter

### DIFF
--- a/apier/v1/attributes_it_test.go
+++ b/apier/v1/attributes_it_test.go
@@ -104,7 +104,7 @@ func TestAttributeSITMongo(t *testing.T) {
 
 func TestAttributeSITInternal(t *testing.T) {
 	alsPrfConfigDIR = "tutinternal"
-	for _, stest := range sTestsAlsPrf {
+	for _, stest := range sTestsAlsPrf[:29] {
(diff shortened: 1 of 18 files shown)


apier/v1/attributes_it_test.go
```diff
@@ -104,7 +104,7 @@ func TestAttributeSITMongo(t *testing.T) {
 
 func TestAttributeSITInternal(t *testing.T) {
 	alsPrfConfigDIR = "tutinternal"
-	for _, stest := range sTestsAlsPrf {
+	for _, stest := range sTestsAlsPrf[:29] {
```
danbogos: I think you forgot the index here

config/cachecfg.go
```diff
@@ -81,3 +81,47 @@ func (cCfg CacheCfg) AsTransCacheConfig() (tcCfg map[string]*ltcache.CacheConfig
 	}
 	return
 }
+
```
danbogos: since this is not something for the user to see, we can add these in an object instead of json string

danbogos:
Thanks!

PR #1737, opened 2019-11-07
Add test for Scheduler filtered

TeoV:


data/conf/samples/filtered_scheduler/cgrates.json
```diff
@@ -0,0 +1,49 @@
...
+
+
+"rals": {
+	"enabled": true,
+},
+
+
+"scheduler": {
+	"enabled": true,
+	"filters": ["*suffix:~AccountID:1001"],
```
danbogos: The filter here should contain also the variable name, in this case should be like: "*suffix:~*act.AccountID:1001"

engine/datamanager.go
```diff
@@ -317,6 +317,25 @@ func (dm *DataManager) CacheDataFromDB(prfx string, ids []string, mustBeCached b
 	return
 }
 
+func (dm *DataManager) GetAccount(id string) (acc *Account, err error) {
+	acc, err = dm.dataDB.GetAccountDrv(id)
+	if err != nil {
```
danbogos: we should only query remote if the error is not found, not for all errors. This is valid for all remote methods

danbogos:
Thanks!

PR #1758, opened 2019-11-18
Added test for SetBalance with Timings

adragusin:


engine/account_test.go
```diff
@@ -2327,6 +2327,53 @@ func TestAccountAsNavigableMap(t *testing.T) {
 	}
 }
 
+func TestAccountClone(t *testing.T) {
+	//empty check
```
danbogos: please take the comments out if not needed

danbogos:
Thanks!

PR #1760, opened 2019-11-19
 Add Destination/ReverseDestination methods in DataManager

TeoV:


engine/datamanager.go
```diff
@@ -869,6 +919,7 @@ func (dm *DataManager) GetActionTriggers(id string, skipCache bool,
 				id, attrs)
 		}
 		if err != nil {
+			err = utils.CastRPCErrToErr(err)
```
danbogos: Please rename to simply CastRPCError

danbogos:
Thanks!

PR #1762, opened 2019-11-20
Implement Replication for all Set Methods

TeoV:
+ Implement RemoveThreshold when replicate is true


utils/coreutils.go
```diff
@@ -1021,7 +1021,7 @@ type SetFilterIndexesArg struct {
 	Indexes      map[string]StringMap
 }
 
-func CastRPCErrToErr(err error) error {
+func CastRPCErr(err error) error {
```
cgrates: please rename to CastRPCError

danbogos:
Thanks!

PR #1785, opened 2019-11-28
Include filters when engine check versions

TeoV:


config/smconfig.go
```diff
@@ -33,6 +33,32 @@ func NewDfltRemoteHost() *RemoteHost {
 	return &dfltVal
 }
 
+type RpcConn struct {
```
danbogos: Please rename to RPCConn and create it's own file

danbogos:
Thanks!

PR #1795, opened 2019-12-02
 Updated DataDB Service. Fixes #1793 

### LINKED ISSUE #1793: data_db needed with FS Agent
Please see:

https://groups.google.com/forum/#!topic/cgrates/oWpEVFFrOzc

```
Hi Gavin,

In last versions we  need access for SessionS to DataDB if you want to
use filters when querying for sessions (optional for some APIs like
GetActiveSessionS). The fact that is mandatory is a BUG, so please open
an issue for that.

In the mean time, simply defining the dataDB should fix it (you can
just add db_type=*internal instead of *redis if you don't want to give
connection to Redis but filters will not work for SessionS).

DanB 
```

Trial97:


danbogos:
Thanks!

PR #1896, opened 2020-01-27
ERs fix for RunDelay and ConcurrentReqs

mintos5:
* small fixes to ESr so we can use it properly with CSV files


mintos5:
Hi @TeoV,
sorry for creating PR before finishing my issue.
I just used **CDRC** as a template for fixing my problems, like here:
https://github.com/cgrates/cgrates/blob/200e51604f24096344271b400c62c05649ba4233/config/cdrccfg.go#L89

And here with **concurrentReqs**:
https://github.com/cgrates/cgrates/blob/200e51604f24096344271b400c62c05649ba4233/cdrc/cdrc.go#L75-L79
Like you can see in old cdrc.go file, the channel was empty initiated to the MaxOpenFiles value.
I can not see this feature in the new filecsv.go

I want to use channel concurrentReqs or conReqs, which is used for example here:
https://github.com/cgrates/cgrates/blob/master/ers/filecsv.go#L114-L117

TeoV:
Hi @mintos5,

Please verify the error because now the code doesn't compile.

Thanks,
TeoV

TeoV:
Hi @mintos5,

Please verify the error and in case this is different than nil return it. 
Something like this : 
if er.RunDelay, err = utils.ParseDurationWithSecs(*jsnCfg.Run_delay);err!=nil{
return
}
This require that *jsnCfg.Run_delay to be a string. Please change the format and also update the tests.
It's not easy to be a CGRateS developer :) 

Thanks,
TeoV

mintos5:
Sorry, I do not know CGRateS source code good enough to fix all the problems relevant to this change. I will close this PR and hopefully, somebody can fix it properly.

Thanks for your patience.

Issue #1859, opened 2020-01-09
Typo in systemd file?


ghenry:
https://github.com/cgrates/cgrates/blob/master/packages/debian/cgrates.service#L19

I don't think there should be that first `-` 

Thanks.

ghenry:
Cool. I can fix! Thanks. Will do.

danbogos:
Hey Gavin,

Sorry, have added my colleague, to the issue and did not realize that I have added you as well. We are planning to completely remove EnvironmentFile directive from .service file, do u see anything wrong with that?

Thanks,
DanB

ghenry:
Hi Dan,

It's nice to use $DAEMON_OPTS from the defaults file. That's the expected
norm on Debian unless you've read otherwise? Working well on Debian 10 for
use which will be here for 5+ years now.

Thanks.

danbogos:
I see, OK, please send us a PR and will check.

Thanks,
DanB

PS: Happy New Year ;).

ghenry:
Will do. Happy New Year all!

sebastic:
> I don't think there should be that first `-`

It's there to not fail when the file does not exist:

> The argument passed should be an absolute filename or wildcard expression, optionally prefixed with "-", which indicates that if the file does not exist, it will not be read and no error or warning message is logged. This option may be specified more than once in which case all specified files are read. If the empty string is assigned to this option, the list of file to read is reset, all prior assignments have no effect.

https://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=

danbogos:
Thanks @sebastic 
Since we had no idea about the reason, we have removed it. It should anyway make not much difference since the file should by default be created by package.

DanB

PS: Nothing against adding it back either.

sebastic:
I would add the "-" back, it's a good practice to make the service more robust. 2/3 of services files in Debian use it for example.

I'm disappointed that the systemd documentation was not consulted to learn why it was used in the first place. It wouldn't have been added in #1618 if it wasn't a good idea.

PR #1912, opened 2020-02-04
Don't fail when /etc/default/cgrates doesn't exist.

sebastic:
It is not a typo as suggested in #1859.

See: https://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=


danbogos:
Thanks!

PR #1929, opened 2020-02-10
Updated FilterS ApierSConns

Trial97:


engine/filters_test.go
```diff
@@ -305,9 +306,37 @@ func TestFilterPassRSRFields(t *testing.T) {
 	}
 }
 
+// structure to mimic the APIerSv1 to test *destination
```
danbogos: Since the RPC is the right way to test this, please cancel this function since it will create duplicate functionality which does not help us in the real tests but giving us fake feeling of working functionality.

danbogos:
Thanks!

PR #1930, opened 2020-02-10
Add backwards compatible with ApierV1 and ApierV2 prefix when calling…

TeoV:
… API's


danbogos:
Please add test for it and also add it to the v0.10 branch also

danbogos:
Thanks!

PR #1935, opened 2020-02-10
Added new ansible script to generate and store a deb package 

Trial97:


data/ansible/generate_package/go.yaml
```diff
@@ -0,0 +1,122 @@
+---
```
danbogos: Please rename folder to something like stable_interim_package

danbogos:
Thanks!

PR #2001, opened 2020-02-24
Updated eventReader default config

### LINKED ISSUE #1994: Bad path value in default config ers->readers->fields->path
Hi CGRateS team, after upgrading to a stable v0.10 version we have found a problem with default config and documentation for ERS service.
The path value is like this:
https://github.com/cgrates/cgrates/blob/646f9a33cdb416d93ace97121dc463bbdf16e195/data/conf/cgrates/cgrates.json#L299
It is missing ***cgreq.** or similar which is needed.
I found correct config in data/conf/samples (I think used for testing)
https://github.com/cgrates/cgrates/blob/646f9a33cdb416d93ace97121dc463bbdf16e195/data/conf/samples/ers_postgres/cgrates.json#L118

Please update the default config and documentation. Thank You

adragusin:
closes #1994


danbogos:
Thanks!

PR #2005, opened 2020-02-28
Correct using a connID with two connections one of it *internal

TeoV:
+ Add support for Mongo for different Marshaler
+In case usage missing from Event for suppliers use 1 Minute as default


danbogos:
Thanks!

PR #2059, opened 2020-03-30
Add verification for event filters before populating data

TeoV:


engine/filters.go
```diff
@@ -77,6 +77,71 @@ func (fS *FilterS) Pass(tenant string, filterIDs []string,
 	return
 }
 
+//verifyPrefixes verify the Element and the Values if has as prefix one of the prefixes
+func verifyPrefixes(rule *FilterRule, prefixes []string) (hasPrefix bool) {
```
danbogos, line 81: Need review together, the following ideas came into my mind:
* The function can be part of the filter object.
* Not sure if additional if for dynamic is needed.
* Can send the prefixes as set instead of slice.
* The use of too many hasPrefix looks a bit odd at least.

As said, need to review it together.

packages/debian/changelog
```diff
@@ -44,6 +44,9 @@ cgrates (0.11.0~dev) UNRELEASED; urgency=medium
   * [SessionS] Move *cost as subflag in *rals for
     SessionSv1.ProcessEvent
   * [DiameterAgent] Added DPR support
+  * [SupplierS] Add verification for event filters before populating
```
danbogos, line 47: Please rename to something like: Prefiltering ~*req. data sources to avoid unnecessary processing of costs/stats.

danbogos:
Thanks!

PR #2144, opened 2020-05-06
Add *monthly timing id support by default ( fixes #2141 )

### LINKED ISSUE #2141: Add monthly recurring timing for ActionPlans Timings
I have a use case for monthly recurring Actions.

TeoV:


danbogos:
Thanks!

PR #2195, opened 2020-05-28
Added non-empty check at migration

adragusin:


danbogos:
Thanks! Please port it to v0.10 also.

PR #2223, opened 2020-06-17
Use correctly SessionTTLUsage when calculate end usage in case of ter…

TeoV:
…minate session from ttl mechanism


danbogos:
Thanks!

PR #2244, opened 2020-06-25
[LoaderS] Add *req as mandatory prefix

TeoV:


ees/virtualEe.go
```diff
@@ -0,0 +1,118 @@
+/*
```
danbogos, line 1: please rename the file to virtualee.go. We don't use caps in our file naming

danbogos:
Thanks!

PR #2274, opened 2020-07-08
Fix panic with configurations with multiple event socket connections (stable branch)

htrendev:
The fix from PR #2273, but for the stable branch.

### DIFF
--- a/agents/fsagent.go
+++ b/agents/fsagent.go
@@ -296,11 +296,11 @@ func (sm *FSsessions) Connect() error {
 		}
 		sm.conns[connIdx] = fSock
 		utils.Logger.Info(fmt.Sprintf("<%s> successfully connected to FreeSWITCH at: <%s>", utils.FreeSWITCHAgent, connCfg.Address))
-		go func() { // Start reading in own goroutine, return on error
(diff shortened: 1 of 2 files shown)


agents/fsagent.go
```diff
@@ -297,7 +297,7 @@ func (sm *FSsessions) Connect() error {
 		sm.conns[connIdx] = fSock
 		utils.Logger.Info(fmt.Sprintf("<%s> successfully connected to FreeSWITCH at: <%s>", utils.FreeSWITCHAgent, connCfg.Address))
 		go func() { // Start reading in own goroutine, return on error
-			if err := sm.conns[connIdx].ReadEvents(); err != nil {
+			if err := fSock.ReadEvents(); err != nil {
 				errChan <- err
 			}
 		}()
```
Trial97, line 303: Can you please add the fsock as a parameter to the goroutine like so:

```suggestion
		go func(fsock *fsock.FSock) { // Start reading in own goroutine, return on error
			if err := fsock.ReadEvents(); err != nil {
				errChan <- err
			}
		}(fSock)
```

htrendev: changed as request

danbogos:
Thanks!

PR #2334, opened 2020-07-31
Added cdrs command in console

Trial97:


config/config_defaults.go
```diff
@@ -69,7 +69,10 @@ const CGRATES_CFG_JSON = `
 	"db_name": "10", 						// data_db database name to connect to
 	"db_user": "cgrates", 					// username to use when connecting to data_db
 	"db_password": "", 						// password to use when connecting to data_db
-	"redis_sentinel":"",					// the name of sentinel when used
+	"redis_sentinel": "",					// the name of sentinel when used
+	"redis_cluster": false,					// if enabled the datadb will try to connect to the redis cluster
```
danbogos, line 73: We need to think of integrating these into uri.

config/config_defaults.go
```diff
@@ -69,7 +69,10 @@ const CGRATES_CFG_JSON = `
...
+	"cluster_sync": "5s",					// the sync interval for the redis cluster
```
danbogos, line 74: Since this applies only to redis, we should consider prefixing for redis.

config/config_defaults.go
```diff
@@ -898,6 +901,10 @@ const CGRATES_CFG_JSON = `
 	"out_datadb_user": "cgrates",
 	"out_datadb_password": "",
 	"out_datadb_encoding" : "msgpack",
+	"out_datadb_redis_sentinel": "",
```
danbogos, line 904: Same story, we should find a way to define them in db uri.

danbogos:
Thanks!

PR #2350, opened 2020-08-18
Update warning logs with correct name ( from SupplierS -> RouteS )

TeoV:


console/statqueue_profile_rem.go
```diff
@@ -22,7 +22,7 @@ import "github.com/cgrates/cgrates/utils"
 
 func init() {
 	c := &CmdRemoveStatQueue{
-		name:      "statqueue_remove",
+		name:      "statqueue_profile_rem",
```
danbogos, line 25: Can rename to stat_profile

console/attribute_profile.go
```diff
@@ -25,7 +25,7 @@ import (
 
 func init() {
 	c := &CmdGetAttributes{
-		name:      "attributes",
+		name:      "attribute_profile",
```
danbogos, line 28: Thinking it should be nicer if we could rename to attributes_profile (like the subsystem). This should apply for all the commands. Need to review together since I am not sure.

danbogos:
Thanks!

Issue #2336, opened 2020-08-03
StatS: Rouding of decimal is not performed in the case of ACD


ryan-netizen:
Greetings, 

i see decimal rounding off from cgrates.json is not taken into account, when calculating ACD for stats, but same applies for ASR. 
Due to no rounding off i see my ACD like this `2.13488372s`

It would be better we if could limit ACD to some value. 

Thanks for the great software. :)

PR #2353, opened 2020-08-19
Add rounding operation for duration metric (e.g. acd, tcd, etc...) fi…

TeoV:
…xes #2336


danbogos:
Thanks!

danbogos:
Zero should be no issue, -1 we can error it as part of config sanitization.

PR #2356, opened 2020-08-20
When use APIerSv1.LoadRatingProfile instead of overwrite combine the RatingPlanActivations

TeoV:


engine/tpreader.go
```diff
@@ -339,7 +339,19 @@ func (tpr *TpReader) LoadRatingProfilesFiltered(qriedRpf *utils.TPRatingProfile)
...
+			if resultRatingProfile, err = tpr.dm.GetRatingProfile(tpRpf.KeyId(), false, utils.NonTransactional); err != nil {
+				if err == utils.ErrNotFound {
+					resultRatingProfile = &RatingProfile{Id: tpRpf.KeyId()}
+					err = nil
+				} else {
+					return err
+				}
+			}
+		} else {
+			resultRatingProfile = &RatingProfile{Id: tpRpf.KeyId()}
```
danbogos, line 352: I would define this as shared variable instead of if ... else.

PR #2377, opened 2020-09-10
DiameterAgent return NOT_FOUND instead of "filter not passing" error …

TeoV:
…and let other subsystem to handle this (e.g. FilterS)


danbogos:
Thanks!

PR #2404, opened 2020-10-02
Add support for options in Elasticsearch Request

TeoV:


utils/consts.go
```diff
@@ -2382,6 +2381,18 @@ const (
 	OptsRouteID = "*routeID"
 	// EEs
 	OptsEEsVerbose = "*eesVerbose"
+	// EEs Elasticsearch options
+	Index               = "index"
```
danbogos, line 2385: Please prefix the options with something like EEs so we know they belong to the same place.

danbogos:
Thanks!

PR #2408, opened 2020-10-05
[AttributeS] Add support for adding fields from other places that event (e.g. Resource.TotalUsage, Stat.MetricName, Account.Balance)

TeoV:


engine/dynamicdp.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	return
+}
+
+func (dDP *dynamicDP) fieldAsInterface(fldPath []string) (val interface{}, err error) {
+	if len(fldPath) < 2 {
+		return nil, fmt.Errorf("invalid fieldname <%s>", fldPath)
+	}
+	switch fldPath[0] {
+	case utils.MetaAccounts:
```
danbogos, line 85: Please add test also for full Account, Resource, Stat. Thanks

danbogos:
Thanks!

PR #2413, opened 2020-10-06
Covered some tests in config directory

porosnicuadrian:
In datdbcfg.go from config directory, there is in error that will never be executed,
https://github.com/cgrates/cgrates/blob/5c558545bda5e92122e5622b3a05de2c2f8dd225/config/datadbcfg.go#L96
 because from this load function 
https://github.com/cgrates/cgrates/blob/5c558545bda5e92122e5622b3a05de2c2f8dd225/config/datadbcfg.go#L184
 there isn't any error to return.


danbogos:
Thanks!

PR #2447, opened 2020-10-20
Modified apiers by adding tenant default value 

porosnicuadrian:
Also moved MinCallDuration,MaxCallDuration from session section to general section and added some tests for rates directory and rateprofile.


danbogos:
Thanks!

PR #2456, opened 2020-10-29
Added AnalyzerSv1.StringQuery API to search over the recorded RPC calls 

Trial97:


analyzers/utils.go
```diff
@@ -0,0 +1,99 @@
+/*
```
danbogos, line 1: Please rename the file to libalalyzers so we can be consistent with other services.

danbogos:
Thanks!

PR #2483, opened 2020-11-13
Covered tests in config directory

porosnicuadrian:
Also, i tried to do some tests on computeRateSIntervals function, but i saved them in my work file.


danbogos:
Thanks!

PR #2491, opened 2020-11-20
Added integration tests for config.go

porosnicuadrian:


config/config.go
```diff
@@ -1561,10 +1561,9 @@ func (cfg *CGRConfig) V1ReloadConfig(args *ConfigReloadArgs, reply *string) (err
 	} else {
 		err = cfg.reloadSections(args.Section)
 	}
-	if err != nil {
-		return
+	if err == nil {
```
danbogos, line 1564: Please keep here the old approach with checking the error first since it is  more clear that we treat it.

PR #2509, opened 2020-12-02
Updated EventCost FieldAsInterface. Fixes #2375

### LINKED ISSUE #2375: [Feature Request] Stats: Calculate metrics on the nested fields of the event
Greetings, 

At the moment i see that metric *tcd calculates the call duration by taking the usage from the top level, but when we charge a customer by 6/6, a call of duration 2 seconds should be counted as 6 seconds in duration. 

I saw that my required value was present inside `Costdetails.Usage` object, so tried to do a sum on that field. 

```
//1 
{
    "_id": ObjectId("5f529ec53f641892e4fb71d7"),
    "id": "ST_12h_Ryan",
    "tpid": "1",
    "activationinterval": {
        "activationtime": "2014-07-29T15:00:00Z",
        "expirytime": ""
    },
    "blocker": false,
    "filterids": [
        "FLTR_rscom"
    ],
    "metrics": [
        {
            "filterids": null,
            "metricid": "*sum:~*req.Costdetails.Usage"
        }
    ],
    "minitems": NumberInt("0"),
    "queuelength": NumberInt("-1"),
    "stored": true,
    "tenant": "cgrates.org",
    "thresholdids": [
        "*none"
    ],
    "ttl": "12h",
    "weight": NumberInt("5")
}
``` 

But looks like this is not supported, gives me below error. 
`<StatQueue> metricID: *sum:~*req.CostDetails.Usage, add eventID: 8b5e34d, error: NOT_FOUND`

It would be nice if we can calculate metrics on nested fields. 

Thanks for the great work. 

Ryan

Trial97:


danbogos:
Thanks!

PR #2542, opened 2020-12-14
Compile the EventCost for sRun before storing it

TeoV:


apier/v1/ees_it_test.go
```diff
@@ -218,7 +218,7 @@ func testEEsExportCDRs(t *testing.T) {
 }
 
 func testEEsVerifyExports(t *testing.T) {
-	time.Sleep(time.Second)
+	time.Sleep(time.Second + 600*time.Millisecond)
```
danbogos, line 221: Please put here directly the final value instead of adding

danbogos:
Thanks!

PR #2591, opened 2021-01-08
Update on AccountProfile from []*Balance to map[string]*Balance

TeoV:


accounts/libaccounts.go
```diff
@@ -29,7 +29,10 @@ import (
 // newAccountBalances constructs accountBalances
 func newAccountBalances(acnt *utils.AccountProfile,
 	fltrS *engine.FilterS, ralsConns []string) (acntBlncs *accountBalances, err error) {
-	blncs := utils.Balances(acnt.Balances) // will be changed with using map so OK to reference for now
+	blncs := make(utils.Balances, len(acnt.Balances)) // Temporary code to pass the build please update this accordingly
+	for _, bal := range acnt.Balances {
+		blncs = append(blncs, bal)
```
danbogos, line 34: Please insert it with index here since it is faster in case of reaching the array size

engine/model_helpers_test.go
```diff
@@ -7227,16 +7227,17 @@ func TestApitoAccountProfileCase2(t *testing.T) {
...
-			Opts: map[string]interface{}{
-				"key1": "val1",
-			},
-		}},
+		Balances: map[string]*utils.Balance{
+			"VoiceBalance": {
+				ID:        "VoiceBalance",
+				FilterIDs: []string{"FLTR_RES_GR2"},
+				Weight:    10,
+				Type:      utils.VOICE,
```
danbogos, line 7235: They type can only be one of *concrete or *abstract. For voice should be *abstract

utils/accountprofile.go
```diff
@@ -128,7 +128,7 @@ func (aP *AccountProfile) Clone() (acnt *AccountProfile) {
 		}
 	}
 	if aP.Balances != nil {
-		acnt.Balances = make([]*Balance, len(aP.Balances))
+		acnt.Balances = make(map[string]*Balance, len(aP.Balances))
 		for i, value := range aP.Balances {
```
danbogos, line 132: As best practice, please use "k" for key and "i" when dealing with slice indexes.

danbogos:
Thanks!

PR #2614, opened 2021-01-18
Add RateProfileIDs to Balance struct and rename from CostAttributes to AttributeIDs ( same as from ChargerProfile )

TeoV:


danbogos:
Thanks!

PR #2719, opened 2021-03-05
Finished config coverage

nickolasdaniel:


config/config_test.go
```diff
@@ -6255,3 +6255,9 @@ func TestCGRConfigGetDP(t *testing.T) {
 		t.Errorf("Expected %+v, received %+v", exp, err)
 	}
 }
+
+func TestLockUnlockSectionsID(t *testing.T) {
```
danbogos, line 6259: Please put it in a function since there is no error here.

danbogos:
Thanks!

PR #2724, opened 2021-03-08
Started testing indexes for every subsystem

porosnicuadrian:


accounts/accounts_test.go
```diff
@@ -56,6 +56,12 @@ func TestListenAndServe(t *testing.T) {
 	accnts.ListenAndServe(stopChan, cfgRld)
 }
 
+func TestShutDownCoverage(t *testing.T) {
```
danbogos, line 59: Please make this as a function since it contains no real tests inside.

danbogos:
Thanks!

PR #2726, opened 2021-03-09
100% Coverage in Registrarc and Tests in ers

andronache98:


registrarc/registrarc_test.go
```diff
@@ -123,3 +123,104 @@ func TestDispatcherHostsService(t *testing.T) {
...
+		cfg:     cfg,
+		connMgr: engine.NewConnManager(cfg, map[string]chan rpcclient.ClientConnector{}),
+	}
+	regist.registerRPCHosts()
+	if !reflect.DeepEqual(regist, registCmp) {
+		t.Errorf("Expected: %+v ,received: %+v", registCmp, regist)
+	}
+}
+
+func TestRegistrarcListenAndServedTmCDispatcher(t *testing.T) {
```
danbogos, line 194: Please rename without Test function since there is no failing test here. Same in other places where we do not test.

PR #2728, opened 2021-03-09
Finished testing all indexes for all subsystems + new update cases

porosnicuadrian:


accounts/accounts_test.go
```diff
@@ -41,7 +41,8 @@ func TestRPCCall(t *testing.T) {
 	}
 }
 
-func TestListenAndServe(t *testing.T) {
+func TestShutDownCoverage(t *testing.T) {
```
danbogos, line 44: If this function has no tests inside, please put it into own function.

PR #2786, opened 2021-04-01
Removed accounts, actions and rates services

Trial97:


apier/v1/accounts_it_test.go
```diff
@@ -1101,7 +1101,7 @@ func testAccITAccountMonthlyEstimated(t *testing.T) {
 		ActionPlan: []*AttrActionPlan{{
 			ActionsId: "TOPUP_ACTION",
 			MonthDays: "31",
-			Time:      "00:00:00",
```
danbogos, line 1104: Need to understand why was here causing issues?

danbogos:
Thanks!

PR #2807, opened 2021-04-07
Coverage tests for config

nickolasdaniel:
Currently checking for coverage in each package. Will add tests where needed.


danbogos:
Thanks!

PR #2852, opened 2021-04-19
Cover funcs in engine

ionutboangiu:


engine/datadbmock.go
```diff
@@ -22,7 +22,11 @@ import (
 	"github.com/cgrates/cgrates/utils"
 )
 
-type DataDBMock struct{}
+type DataDBMock struct {
```
danbogos, line 25: Do we need mock for DataDB? Not possible to use internalDB?

PR #2887, opened 2021-04-29
Implemented *datetime type in AgentRequest and EventRequest and added…

andronache98:
… tests for them


danbogos:
Thanks! Please port it to 1.0 also.

PR #2922, opened 2021-05-10
Tested ReloadCache function in engine/tpreader.go 

nickolasdaniel:
Also, added missing Timing prefix in GetLoadedIds and Timing id in ReloadCache


danbogos:
Thanks!

PR #2926, opened 2021-05-10
Cover funcs in engine

ionutboangiu:


engine/resources.go
```diff
@@ -289,9 +289,7 @@ func (rs Resources) allocateResource(ru *ResourceUsage, dryRun bool) (alcMessage
 		for _, r := range rs {
 			r.removeExpiredUnits()
 			if _, hasID := r.Usages[ru.ID]; hasID && !dryRun { // update
-				if err = r.clearUsage(ru.ID); err != nil {
-					return
-				}
+				r.clearUsage(ru.ID)
```
danbogos, line 292: Please add comment with the reason for ignoring error herre.

PR #2938, opened 2021-05-11
Worked for RateProfileCost and RateSIntervals

porosnicuadrian:


danbogos:
Please review the code since Rates needs to be map[string]*IntervalRate, otherwise we will not perform any compress.
https://github.com/cgrates/cgrates/blob/23e1a4c84b7cd34388f9181ae237a6ab3c749c49/utils/librates.go#L537

PR #2942, opened 2021-05-12
RateProfileCost and Equals/As methods changes + tests

porosnicuadrian:


rates/rates.go
```diff
@@ -195,12 +195,17 @@ func (rS *RateS) rateProfileCostForEvent(ctx *context.Context, rtPfl *utils.Rate
 	if ivalStart, err = args.IntervalStart(); err != nil {
 		return
 	}
-	if rpCost.RateSIntervals, err = computeRateSIntervals(ordRts, ivalStart, usage); err != nil {
+	rpCost.Rates = make(map[string]*utils.IntervalRate)
```
danbogos, line 198: Please initialize the map when creating rpCost so we have things centralized.

rates/rates_test.go
```diff
@@ -175,17 +174,24 @@ func TestRateProfileCostForEvent(t *testing.T) {
 				Increments: []*utils.RateSIncrement{
 					{
 						IncrementStart:    utils.NewDecimal(0, 0),
-						Rate:              rPrf.Rates["RATE1"],
 						IntervalRateIndex: 0,
+						RateID:            "RATE1",
```
danbogos, line 178: In order to keep things close to reality here, please use UUID instead of clear IDs since it might confuse us in the future.

utils/eventcharges.go
```diff
@@ -364,10 +380,11 @@ type ExtEventCharges struct {
...
 	Charges []*ChargeEntry
 
-	Accounting  map[string]*ExtAccountCharge
-	UnitFactors map[string]*ExtUnitFactor
-	Rating      map[string]*ExtRateSInterval
-	Accounts    map[string]*ExtAccount
+	Accounting    map[string]*ExtAccountCharge
+	UnitFactors   map[string]*ExtUnitFactor
+	Rating        map[string]*ExtRateSInterval
+	IntervalRates map[string]*ExtIntervalRate
```
danbogos, line 386: Please rename to Rates instead of IntervalRates, so we can be in sync with RateProfileCost.

utils/librates.go
```diff
@@ -577,27 +711,34 @@ func (rIv *RateSInterval) CompressEquals(rIv2 *RateSInterval) (eq bool) {
...
 func (rIcr *RateSIncrement) CompressEquals(rIcr2 *RateSIncrement) (eq bool) {
-	return rIcr.Rate.UID() == rIcr2.Rate.UID() &&
+	return rIcr.RateID == rIcr2.RateID &&
 		rIcr.IntervalRateIndex == rIcr2.IntervalRateIndex &&
 		rIcr.Usage.Big.Cmp(rIcr2.Usage.Big) == 0
 }
 
 // Cost computes the Cost on RateSIncrement
-func (rIcr *RateSIncrement) Cost() *decimal.Big {
+func (rIcr *RateSIncrement) Cost(rts map[string]*IntervalRate) (cost *decimal.Big, err error) {
```
danbogos, line 736: Please remove the error here since panic should be enough in case of wrong populating of the map.

danbogos:
Thanks!

PR #2993, opened 2021-05-26
Added unit tests for config

nickolasdaniel:
Also renamed config_apis.go to apis.go, and config_apis_test.go to apis_test.go
Switched from "&" to new() in apis.go


danbogos:
Thanks!

PR #3013, opened 2021-05-28
Updated check config sanity. Fixes #3008 

### LINKED ISSUE #3008: ERs: amqp reader for the cdrs
Greetings, 

From the test section I see that cgrates can read cdrs from the amqp queue. 

https://github.com/cgrates/cgrates/blob/master/ers/amqp_it_test.go#L42

But when I try to use that example actually, i get below error back, 
```
Could not parse config: <<ERs> unsupported data type: *amqp_json_map for reader with ID: amqp>
```

This seems to be coming from below section, 
https://github.com/cgrates/cgrates/blob/master/config/config.go#L321

It would be really nice, if we can also read cdrs from the amqp and bill them like we do like the case of csv and json files. 

In opensips you can send cdrs to rabbitmq and i would like to bill them in the cgrates instead of using the csv files for the billing (which does not seem to fit for the realtime cdrs). 

Below is an example of CDR request posted to rabbitmq queue by opensips.  

```
{
  "jsonrpc": "2.0",
  "method": "E_ACC_CDR",
  "params": {
    "method": "INVITE",
    "from_tag": "jF93400eQmtSg",
    "account": "12345",
    "subject": "12345",
     "category":"*call",
    "to_tag": "2Ftp805pa5y4j",
    "callid": "f6e266cf-3984-123a-e3b8-56000192591d",
    "sip_code": "200",
    "sip_reason": "OK",
    "time": 1622116396,
    "duration": 11,
    "ms_duration": 10512,
    "setuptime": 0,
    "created": 1622116396
  }
}
```

Thanks again for the awesome work, you guys are doing. 

Ryan

Trial97:


data/storage/mysql/create_db_with_users.sql
```diff
@@ -7,3 +7,9 @@ DROP DATABASE IF EXISTS cgrates;
 CREATE DATABASE cgrates;
 
 GRANT ALL on cgrates.* TO 'cgrates'@'localhost' IDENTIFIED BY 'CGRateS.org';
+
+-- extra DB for ees and ers
+DROP DATABASE IF EXISTS cgrates2;
```
danbogos, line 12: Please  move this part in a separate script.

danbogos:
Thanks!

PR #3032, opened 2021-06-10
Rename dispatcher to dispatchers in registrarc and Changed dispatcher in registrarc to dispatchers

andronache98:


danbogos:
Thanks!

PR #3090, opened 2021-06-29
Testing console commands in cmd/cgr-console

nickolasdaniel:
There are also some functions which require a different engine config and as well as a different tariff plan in order to be tested.


danbogos:
Thanks!

PR #3105, opened 2021-07-01
Testing console commands in cmd/cgr-console

nickolasdaniel:
Tested all available commands, and created a specific config (data/conf/samples/loaders/loaders_console_mongo, data/conf/samples/loaders/loaders_console_mongo) in order to test commands which are related to loaders.
Some of the tests need to be looked up again as they are not always succeding.


danbogos:
Thanks!

PR #3142, opened 2021-07-12
Add tests for *sipcid case from ParseAttribute function

ionutboangiu:


ghenry:
Was this because of my question to the mailing? :)

danbogos:
> Was this because of my question to the mailing? :)

Which question please?

PR #3146, opened 2021-07-12
Tested V1CostForEvent with both values of RateIndexedSelects 

porosnicuadrian:


danbogos:
Thanks!

rates/rates.go
```diff
@@ -128,7 +128,7 @@ func (rS *RateS) rateProfileCostForEvent(ctx *context.Context, rtPfl *utils.Rate
 		utils.MetaReq:  args.CGREvent.Event,
 		utils.MetaOpts: args.APIOpts,
 	}
-	var rtIDs utils.StringSet
+	rtIDs := utils.StringSet{}
```
danbogos, line 131: Is therre a specific reason for this init here? Thanks!

porosnicuadrian: Did this to avoid panic vrom populating that StringSet if  RateIndexedSelect is false

PR #3152, opened 2021-07-14
Added extra fields for RemoteHost and DispatcherHost related structs as well as redesigned functions where needed

nickolasdaniel:


danbogos:
Thanks!

PR #3153, opened 2021-07-14
Added extra fields for RemoteHost and DispatchHost related struct as well as redesigned functions where needed

nickolasdaniel:


danbogos:
Thanks!

PR #3269, opened 2021-08-31
ers docs typo

anthonysomerset:
references content_fields instead of fields which can cause an issue - this also needs to be fixed in the 0.10.x branch as well potentially, was unsure of process for a fix outside the master branch


danbogos:
Thanks!

PR #3290, opened 2021-09-14
 Added support for in memory db for analyzers 

Trial97:


analyzers/analyzers.go
```diff
@@ -61,6 +61,10 @@ func (aS *AnalyzerService) SetFilterS(fS *engine.FilterS) {
 }
 
 func (aS *AnalyzerService) initDB() (err error) {
+	if aS.cfg.AnalyzerSCfg().DBPath == utils.EmptyString {
```
danbogos, line 64: Please make this working with "*internal" instead of empty, so it does not look like a patch.

Thanks!

danbogos:
Thanks!

PR #3303, opened 2021-09-21
Replace OptsRatesUsage with OptsAccountsUsage where necessary and ref…

ionutboangiu:
…actor OptsAsDecimal


utils/consts.go
```diff
@@ -2195,6 +2195,7 @@ const (
 	OptsContext                = "*context"
 	Subsys                     = "*subsys"
 	MetaMethod                 = "*reqMethod"
+	OptsAccountsUsage          = "*accountsUsage"
```
danbogos, line 2198: In order to be consistent, you need to keep also account options (as all the other) prefixed with 3-4 chars instead of full accounts or rates.

danbogos:
Thanks!

PR #3324, opened 2021-09-27
Add accounts and rates conns to session's cfg section and change type of thresholds opts to map

ionutboangiu:


danbogos:
Thanks!

PR #3344, opened 2021-10-04
1.0

porosnicuadrian:


sessions/sessions.go
```diff
@@ -1583,10 +1585,12 @@ func (sS *SessionS) BiRPCv1AuthorizeEvent(ctx *context.Context,
 		args.APIOpts = make(map[string]interface{})
 	}
 	if v, has := args.APIOpts[utils.OptsAttributeS]; !has {
-		if attrS, err = engine.FilterBoolCfgOpts(ctx, args.Tenant, args.CGREvent.AsDataProvider(), sS.filterS,
-			sS.cgrCfg.CdrsCfg().Opts.Attributes); err != nil {
-			return
-		}
+		/*
+			if attrS, err = engine.FilterBoolCfgOpts(ctx, args.Tenant, args.CGREvent.AsDataProvider(), sS.filterS,
```
danbogos, line 1589: Are you disabling here the dynamic filters all gogether?

danbogos:
Thanks!

sessions/sessions.go
porosnicuadrian: I disabled those cases for now because these are taken from sS.cgrCfg.CdrsCfg().Opts(CDRs config), and it should be implementer for sessions config.

PR #3402, opened 2021-10-20
New method for CGREngine for getting dm

porosnicuadrian:


ees/rpc.go
```diff
@@ -62,8 +62,13 @@ func (e *RPCee) Cfg() (eCfg *config.EventExporterCfg) {
 
 func (e *RPCee) Connect() (err error) {
 	e.Lock()
-	if e.conn, err = rpcclient.NewRPCClient(context.TODO(), utils.TCP, e.cfg.ExportPath, e.tls, e.keyPath, e.certPath, e.caPath, 1, 1, e.connTimeout, e.replyTimeout, e.codec, nil, false, nil); err != nil {
-		return err
+	if e.conn != nil {
```
danbogos, line 65: Not sure I understand this. Could it be wrongly placed?

services/cgr-engine.go
```diff
@@ -119,6 +119,10 @@ func (cgr *CGREngine) GetServDeps() map[string]*sync.WaitGroup {
 	return cgr.srvDep
 }
 
+func (cgr *CGREngine) GetDbService() *DataDBService {
```
danbogos, line 122: Do we still need this method now that we are not longer using dataDB outside of engine?

PR #3474, opened 2021-11-26
Added new api for exporter -ArchiveEventsAsReply + test

porosnicuadrian:


ees/ee_test.go
```diff
@@ -82,11 +83,11 @@ func TestNewEventExporterCase2(t *testing.T) {
 	if err != nil {
 		t.Error(err)
 	}
-	eeExpect, err := NewFileFWVee(cgrCfg.EEsCfg().Exporters[0], cgrCfg, filterS, dc)
+	eeExpect, err := NewFileFWVee(cgrCfg.EEsCfg().Exporters[0], cgrCfg, filterS, dc, io.Discard)
```
danbogos, line 86: In order to init the buffer properly, should we not use nil instead of io.Discard?

danbogos:
Thanks!

PR #3509, opened 2021-12-22
Updated rsyslog files

Trial97:


sessions/sessions.go
```diff
@@ -422,28 +422,30 @@ func (sS *SessionS) debitSession(s *Session, sRunIdx int, dur time.Duration,
 	}
 	sr := s.SRuns[sRunIdx]
 	rDur := sr.debitReserve(dur, lastUsed) // debit out of reserve, rDur is still to be debited
+	fmt.Println("rDur ", rDur)
```
danbogos, line 425: Please remove the prints.

sessions/sessions.go
```diff
@@ -861,9 +863,9 @@ func (sS *SessionS) getIndexedFilters(tenant string, fltrs []string) (
 		f, err := engine.GetFilter(sS.dm, tenant, fltrID,
 			true, true, utils.NonTransactional)
 		if err != nil {
-			if err == utils.ErrNotFound {
-				err = utils.ErrPrefixNotFound(fltrID)
-			}
+			// if err == utils.ErrNotFound {
```
danbogos, line 866: Here we need to at least log it, continue will not do since there can be more serious errors coming from the database.

danbogos: Best is still to return it.

danbogos:
Thanks!

PR #3515, opened 2022-01-03
Implemented prometheus trough stats

porosnicuadrian:


config/config.go
```diff
@@ -187,6 +187,7 @@ func newCGRConfig(config []byte) (cfg *CGRConfig, err error) {
 			ProfileIDs:           []*utils.DynamicStringSliceOpt{},
 			ProfileIgnoreFilters: []*utils.DynamicBoolOpt{},
 			RoundingDecimals:     []*utils.DynamicIntOpt{},
+			PrometheusMetrics:    []*utils.DynamicStringSliceOpt{},
```
danbogos, line 190: Need some better naming here since in the end are stat profiles and not really metrics we define.

danbogos:
Thanks!

PR #3554, opened 2022-01-25
Replaced CGRID with originID

andronache98:


danbogos:
Please add the *originID in the APIOpts as agreed in the ticket. Thanks!

PR #3560, opened 2022-01-28
Changed CGRID to originID and moved it to opts and changed Weight in Weights with DynamicsWeights value in Attributes

andronache98:


danbogos:
Please review the whole PR and make sure the *originID is in the APIOpts

PR #3562, opened 2022-02-02
Changed the build script and GetCGRVersion() 

nickolasdaniel:
So that we get CommitDate instead of AuthorDate when versioning the engine.


danbogos:
Please add the right test for this modification. Thanks!

danbogos:
Thanks!

PR #3572, opened 2022-02-09
DynamicWeights in ActionProfile

andronache98:


utils/apitpdata.go
```diff
@@ -654,7 +654,7 @@ type TPActionProfile struct {
 	Tenant    string
 	ID        string
 	FilterIDs []string
-	Weight    float64
+	Weights   string
```
danbogos, line 657: Not sure about this one. Why do we have it here as string?

danbogos:
Thanks!

PR #3577, opened 2022-02-10
Remove refund sequence from store block (CDRsProcessEvent) and make C…

ionutboangiu:
…ostDetails nil after a refund


danbogos:
Thanks!

PR #3587, opened 2022-02-24
Combined matchedProfile and alteredField for attributes

porosnicuadrian:


agents/astagent.go
```diff
@@ -182,20 +182,22 @@ func (sma *AsteriskAgent) handleStasisStart(ev *SMAsteriskEvent) {
...
+					continue //maybe removed
+				}
+				fldVal, err := authReply.Attributes.CGREvent.FieldAsString(fldName)
+				if err != nil {
+					utils.Logger.Warning(
+						fmt.Sprintf(
+							"<%s> error <%s> extracting attribute field: <%s>",
+							utils.AsteriskAgent, err.Error(), fldName))
+				}
+				if !sma.setChannelVar(ev.ChannelID(), fldName, fldVal) {
```
danbogos, line 198: we need to process altered in a StringSet to make sure we do not set duplicates in agents.

agents/diamagent_test.go
```diff
@@ -259,8 +263,12 @@ func TestProcessRequest(t *testing.T) {
 			}
 			*prply = sessions.V1UpdateSessionReply{
 				Attributes: &engine.AttrSProcessEventReply{
-					MatchedProfiles: []string{"ATTR_1001_SESSIONAUTH"},
-					AlteredFields:   []string{"*req.Password", "*req.PaypalAccount", "*req.RequestType", "*req.LCRProfile"},
+					AlteredFields: []*engine.FieldsAltered{
+						{
+							MatchedProfileID: "ATTR_1001_SESSIONAUTH",
+							AlteredFields:    []string{"*req.Password", "*req.PaypalAccount", "*req.RequestType", "*req.LCRProfile"},
```
danbogos, line 269: Please rename AlteredFields here into Fields, as agreed since we have a duplication.

agents/astagent.go
danbogos: Maybe a method on the structure, returning unique altered fields.

danbogos:
Thanks!

PR #3657, opened 2022-04-20
Add timestamp to exported logs

ionutboangiu:


engine/logger.go
```diff
@@ -69,6 +71,7 @@ func (el *ExportLogger) call(m string, level int) error {
 				utils.NodeID: el.nodeID,
 				"Message":    m,
 				"Severity":   level,
+				"Timestamp":  timestamp.Format("2006-01-02 15:04:05"),
```
danbogos, line 74: Please keep here the standard timestamp generated by Go. In your format I think the zimezone will be stripped.

Thanks!

PR #3659, opened 2022-04-21
Blockers for attributes and configs

porosnicuadrian:


utils/apitpdata.go
```diff
@@ -319,7 +319,7 @@ type TPAttributeProfile struct {
 	ID         string
 	FilterIDs  []string
 	Attributes []*TPAttribute
-	Blocker    bool
+	Blockers   string
```
danbogos, line 322: We might need here proper structure.

danbogos:
Thanks!

Issue #3705, opened 2022-09-07
Debian repo key has expired


gedia:
Hello,

The GPG key used to sign the cgrates debian repository seems to have expired as of 2022-08-14. Would you consider extending it or replacing it so that we don't have to explicitly trust the repo? Thanks!

ionutboangiu:
Hi @gedia,

We have generated a new GPG key.  Before updating the apt packages, please add the new key by using the following command: `sudo wget -O - https://apt.cgrates.org/apt.cgrates.org.gpg.key | sudo apt-key add -`

Thanks,
Ionuț

sebastic:
`apt-key` is deprecated:
> Use of `apt-key` is deprecated, except for the use of `apt-key del` in maintainer scripts to remove existing keys from the main keyring. If such usage of `apt-key` is desired the additional installation of the GNU Privacy Guard suite (packaged in gnupg) is required.
> 
> [`apt-key(8)`](https://manpages.debian.org/bullseye/apt/apt-key.8.en.html) will last be available in Debian 11 and Ubuntu 22.04.
> 
> [...]
> 
> _Note:_ Instead of using this command a keyring should be placed directly in the `/etc/apt/trusted.gpg.d/` directory with a descriptive name and either "gpg" or "asc" as file extension.

I.e.:
```
sudo wget https://apt.cgrates.org/apt.cgrates.org.gpg.key -O /etc/apt/trusted.gpg.d/cgrates.asc
```

ionutboangiu:
Thanks for your contribution!

PR #3711, opened 2022-09-14
Use /etc/apt/trusted.gpg.d instead of apt-key.

sebastic:
As reported in https://github.com/cgrates/cgrates/issues/3705#issuecomment-1246264698, `apt-key` is deprecated, `/etc/apt/trusted.gpg.d` should be used instead.


danbogos:
Thanks!

PR #3715, opened 2022-09-16
Disaptcher opts tests

porosnicuadrian:


general_tests/dispatcher_opts_it_test.go
```diff
@@ -114,14 +116,17 @@ func testDispatcherOptsAdminSetDispatcherProfile(t *testing.T) {
 			Tenant: "cgrates.org",
 			RemoteHost: &config.RemoteHost{
 				ID:              "SELF_ENGINE",
-				Address:         "*localhost",
+				Address:         "*internal",
```
danbogos, line 119: As discussed in the past, please keep here the localhost since we can see it better when troubleshooting.

Thanks!

danbogos:
Thanks!

PR #3719, opened 2022-09-21
Dispatcher test + dp parameter on GetBoolOpts

porosnicuadrian:


engine/libdynopts.go
```diff
@@ -171,19 +171,19 @@ func GetIntOpts(ctx *context.Context, tnt string, ev *utils.CGREvent, fS *Filter
 
 // GetBoolOpts checks the specified option names in order among the keys in APIOpts returning the first value it finds as bool, otherwise it
 // returns the config option if at least one filter passes or the default value if none of them do
-func GetBoolOpts(ctx *context.Context, tnt string, ev *utils.CGREvent, fS *FilterS, dynOpts []*utils.DynamicBoolOpt,
+func GetBoolOpts(ctx *context.Context, tnt string, dP utils.DataProvider, fS *FilterS, dynOpts []*utils.DynamicBoolOpt,
 	dftOpt bool, optNames ...string) (cfgOpt bool, err error) {
+	optsDP := GetAPIOptsFromDataProvider(dP)
 	for _, optName := range optNames {
-		if opt, has := ev.APIOpts[optName]; has {
+		if opt, has := optsDP[optName]; has {
```
danbogos, line 178: here is a source of bugs since if the optsDP is nil, you will get panic.

engine/libengine.go
```diff
@@ -182,6 +182,19 @@ func NewServiceWithName(val interface{}, name string, useName bool) (_ IntServic
 	return s, nil
 }
 
+func GetAPIOptsFromDataProvider(dP utils.DataProvider) map[string]interface{} {
```
danbogos, line 185: Data provider is not an map, please use the data provider methods here.

danbogos: Since the function is not used anymore, please remove it.

Thanks!

danbogos:
Thanks!

PR #3756, opened 2022-10-29
Added check for explicit empty values in mailer config to allow SMTP without auth.

ruhnet:
If empty values are provided in the config for user/password, SMTP authentication will not be used when sending mails. This is useful when the mail server being utilized is localhost (or on a trusted network) and doesn't use authentication.


danbogos:
Thanks!

PR #3846, opened 2023-01-18
Use dpkg-buildpackage --no-sign when the GPG key is not available.

sebastic:
The `--no-sign` option for `dpkg-buildpackage` needs to be used for package builds by others.

This change needs to be backported to v0.10 too.


danbogos:
Thanks  @sebastic !

@gezimbll, please add this to v0.10 and 1.0 branches as well.

PR #3888, opened 2023-02-23
Improve logger tests

arberkatellari:


utils/logger.go
```diff
@@ -34,7 +34,8 @@ func init() {
 
 		var err error
 		//used for testing only, so we will ignore the error for now
-		if Logger, err = Newlogger(MetaSysLog, EmptyString); err != nil {
+		Logger, err = Newlogger(MetaSysLog, EmptyString)
```
danbogos, line 37: Please integrate here the if, as it was since it is shared.

danbogos:
Thanks!

utils/logger.go
arberkatellari: Done in this commit: [https://github.com/cgrates/cgrates/commit/e7034971e8da750b68eb085ebb08e2e455f96f78](https://github.com/cgrates/cgrates/commit/e7034971e8da750b68eb085ebb08e2e455f96f78)

PR #3930, opened 2023-04-19
Revise CDR rerating

ionutboangiu:
If the reRate parameter is set to true, also set the refund to true.

In case CostDetails is not populated, before attempting to refund try to retrieve it from StorDB.

Now that the refund happens before the debit, revise the expected values for the testV1CDRsProcessEventWithRefund subtest within the apier/v1/cdrs_it_test.go file.

### DIFF
--- a/engine/cdrs.go
+++ b/engine/cdrs.go
@@ -439,6 +439,9 @@ func (cdrS *CDRServer) eeSProcessEvent(cgrEv *CGREventWithEeIDs) (err error) {
 // in case of partially executed, both error and evs will be returned
 func (cdrS *CDRServer) processEvents(evs []*utils.CGREvent,
 	chrgS, attrS, refund, ralS, store, reRate, export, thdS, stS bool) (outEvs []*utils.EventWithFlags, err error) {
+	if reRate {
+		refund = true
+	}
 	if attrS {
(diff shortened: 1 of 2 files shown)


engine/cdrs.go
```diff
@@ -513,6 +516,18 @@ func (cdrS *CDRServer) processEvents(evs []*utils.CGREvent,
...
 		for i, cdr := range cdrs {
+			if cdr.CostDetails == nil { // if CostDetails is not populated, look for it inside the previously stored cdr
+				var prevCDRs []*CDR // only one should be returned
+				if prevCDRs, _, err = cdrS.cdrDb.GetCDRs(
+					&utils.CDRsFilter{CGRIDs: []string{cdr.CGRID},
+						RunIDs: []string{cdr.RunID}}, false); err != nil {
+					utils.Logger.Info(
+						fmt.Sprintf("<%s> could not retrieve previously stored CDR, error: <%s>",
+							utils.CDRs, err.Error()))
+					continue
```
danbogos, line 527: Here we should return PartiallyExecuted and write error on log since we cannot refund if we cannot find previous CostDetails.

danbogos:
Thanks!

PR #3933, opened 2023-04-19
Revise CDR rerating

ionutboangiu:
If the reRate parameter is set to true, also set the refund to true.

In case CostDetails is not populated, retrieve it from StorDB if possible and add it to the CGREvent before converting to CDRs.

Now that the refund happens before the debit, revise the expected values for the testV1CDRsProcessEventWithRefund subtest within the apier/v1/cdrs_it_test.go file.


danbogos:
Thanks!

PR #3934, opened 2023-04-19
Revise CDR rerating

ionutboangiu:
The (*CDRServer).processEvent function is now called processEvents and can be passed an array of CGREvents instead of only one. This was done because when calling the RateCDRs API we want to first refund all CDRs before starting to debit again.

Now, the refundEventCost function returns an additional boolean, that signals whether the refund occured or didn't.

If the reRate parameter is set to true, also set refund to true.

In case CostDetails is not populated, retrieve it from StorDB if possible and add it to the CGREvent before converting to CDRs. Set CostDetails back to nil once the refund goes through.

Remove the refund logic from within the store block.

Now that the refund happens before the debit, revise the expected values for the "testV1CDRsProcessEventWithRefund" subtest within the apier/v1/cdrs_it_test.go file.

Add an integration test for the following scenario:
 -create one account with one balance of 1 free minute and rating for the rest.
 -send one CDR of two minutes with ProcessEvent. This should consume 60s out of
the free balance and charge 60s. The SetupTime in the CDR should be 1 hour after the second CDR.
 -send the second CDR with an usage of 2m. This should be charged entirely.
 -send a RateCDR API call with OrderBy: "SetupTime". This should rerate the two
CDRs from above and change their order of rating.

### DIFF
--- a/apier/v1/cdrs_it_test.go
+++ b/apier/v1/cdrs_it_test.go
@@ -241,9 +241,9 @@ func testV1CDRsProcessEventWithRefund(t *testing.T) {
 	}
 	if err := cdrsRpc.Call(utils.APIerSv2GetAccount, acntAttrs, &acnt); err != nil {
 		t.Error(err)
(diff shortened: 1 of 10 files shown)


danbogos:
Thanks!

PR #3956, opened 2023-05-19
Improve cost formatting and parsing for CDRs

ionutboangiu:
The FormatCost function in cdr.go now accepts an additional parameter of type *config.RSRParser. This is then used to extract the value from its path as opposed to always using the value of the Cost field directly.

Improved the unit test for the FormatCost function. Now it has become a table-driven test and it handles cases when the cost is retrieved from different fields other than  from the CDR.


danbogos:
Thanks!

PR #3964, opened 2023-05-22
Revise err handling for CSV storage constructor

ionutboangiu:
NewFileCSVStorage() now returns an error besides the storage struct itself, which is logged and returned instead of calling log.Fatal() which was causing the engine to crash.

Fixed compilation errors by creating the CSVStorage separately and passing it as an argument to the TpReader constructor.


danbogos:
Thanks!

PR #3981, opened 2023-05-25
Add error handler for Mongo GetCDrs remove

arberkatellari:


danbogos:
Thanks!

engine/storage_mongo_stordb.go
```diff
@@ -1100,6 +1100,9 @@ func (ms *MongoStorage) GetCDRs(qryFltr *utils.CDRsFilter, remove bool) ([]*CDR,
 		var chgd int64
 		err := ms.query(func(sctx mongo.SessionContext) (err error) {
```
danbogos, line 1101: Here we could make it nicer and use variable overload, by defining it in the return signature and use it here

PR #4012, opened 2023-06-22
Fix message on agents logs

gezimbll:


agents/libagents.go
```diff
@@ -52,17 +53,17 @@ func processRequest(ctx *context.Context, reqProcessor *config.RequestProcessor,
 	}
 	if reqProcessor.Flags.Has(utils.MetaLog) {
 		utils.Logger.Info(
-			fmt.Sprintf("<%s> LOG, processorID: %s, diameter message: %s",
-				agentName, reqProcessor.ID, agReq.Request.String()))
+			fmt.Sprintf("<%s> LOG, processorID: %s, %s message: %s",
+				agentName, reqProcessor.ID, strings.ToLower(agentName[:len(agentName)-5]), agReq.Request.String()))
```
danbogos, line 57: Please do not use toLower as agent name. All agents should be presented with their full name, ie: DNSAgent in this case. If already done, please modify everywhere.

danbogos:
Thanks!

PR #4028, opened 2023-06-30
Upgrade MongoDB driver to v1.12 and implement nginx&gpg ansible roles

ionutboangiu:
- Set (but comment) serverAPI options (currently distinct api and
create.size BSON field are deprecated + possible others that are untested)
- Rename and refactor 'TimeDecodeValue1' to 'decodeTimeValue', add
nil case handling
- Implement 'buildURL' function to connect to mongo (can also be
used for mysql and postgres)
- Update function names, variable names, and comments for clarity
- Replace 'bsonx.Regex' with the Regex primitive for v1.12 compatibility
- Use simple concatenation instead of Sprintf
- Declare 'timeType' locally, replace global 'tTime'
- Use 'RegisterTypeDecoder' to replace deprecated 'RegisterDecoder'
- Simplify several functions without altering functionality
- Remove redundant 'storageType' field in 'mongoStorage',
use 'isDataDB' instead


danbogos:
Thanks!

PR #4045, opened 2023-07-13
Added filter implementation for sentrypeer

gezimbll:


ghenry:
Hi @gezimbll

Thanks for doing this!

A couple of issues which I hope are opportunities for me to improve our docs (docs.sentrypeer.com):

1. SentryPeer is different from APIBAN, but you've appended to https://github.com/cgrates/cgrates/blob/600b49488ff8ef4857c40eb8be3b37daf99bd203/config/apiban.go

2. The authentication flow needs to grab the access token first. So cfg needs to store the client_id and client_secret. Grab token and store in cache on first GET. 

The token used as a bearer tokenis valid for 1 day. Therefore, if you get a 40X reply on API GETs, try to get a fresh access token and do the GET again. Then give up of still an issue. Do not request a fresh token every GET. 

https://docs.sentrypeer.com/authentication

The rest looks good!

ghenry:
Here are two examples of you'd like some code to read:

1. https://github.com/SentryPeer/SentryPeer-FreePBX-Module/blob/d99985d51d87eda3f27e728cf83d28f9b7d4d418/Sentrypeer.class.php#L165C36-L165C36

https://github.com/SentryPeer/SentryPeer-FreePBX-Module/blob/d99985d51d87eda3f27e728cf83d28f9b7d4d418/agi-bin/sentrypeer.php#L33

2. https://github.com/SentryPeer/SentryPeer/blob/02173c8c09b89f7692ac5252a48d1eee9b785429/src/json_logger.c#L617

Thanks!

danbogos:
Thanks!

ghenry:
Amazing work! Will have a review and play. Thank you again!

What's the best way to link to this integration from my docs etc?

PR #4085, opened 2023-08-29
Activate privilege escalation for removing go installation

ionutboangiu:
To remove files from /usr/local/go, root privileges are required.


danbogos:
Thanks!

PR #4109, opened 2023-09-24
Upgrade NATS driver version and implementation

ionutboangiu:
Upgraded NATS server version in ansible role.

upgraded go.mod nats version due to an issue caused by version mismatch between driver and server (uncertain).

Increased amount of sleep in tests after starting a NATS server with jetstream enabled from 50ms to 100ms.

## *NatsER.Serve

- Replaced ChanQueueSubscribe with QueueSubscribe for Core NATS consumer to handle the message processing directly.

- Since QueueSubscribe is now used regardless of jetstream status, the message handler has been assigned to a separate variable that can be reused.

-  The message handler is now dealing with the message processing directly, therefore the select case listening for the channel which is feeding NATS messages can be removed together with the channel itself and the select. Currently, the goroutine within Serve only has to block until the rdrExit chan is closed.

- Moved the resource check inside the handler right before starting the message processing goroutine.

## ers.getProcessedOptions

- Renamed function from getProcessOptions to getProcessedOptions.

- Initially, the EventExporterOpts struct was always initialized to be non-nil which meant exporting processed messages by the reader was always enabled by force. Function has been updated to initialize it only once when the first Processed option that was set is found.

- Created init function for all types of opts structs to avoid repetition.

## *NatsEE.parseOpts

- Renamed function from parseOpt to parseOpts.

- Updated function to return early in case of nil opts struct to reduce nesting.

- The nested jetstream status and maxwait conditions nil verification have been merged into one condition.

- Handled the error coming from GetNatsOpts function.

- NATS Subject assignment has been removed. It was redundant, since it had already been set from before this function was called.

## *NatsEE.Connect

- Updated function to return early in case of non-nil nats.Conn value to reduce nesting.

## *NatsEE.ExportEvent

- Use defer to release resources and RUnlock.

## *NatsEE.Close

- Use defer to Unlock.

- Update function to return early in case of nil nats.Conn value to reduce nesting.

## ees.GetNatsOpts

- Chose switch over if else when parsing client certificate and keys opts.

- Updated function to return the errors directly instead of assigning them to a separate variable right before returning.

## ers.GetNatsOpts

- Passed the NATSROpts struct directly to the function.

- Chose switch over if else when parsing client certificate and keys opts.

- Updated function to return the errors directly instead of assigning them to a separate variable right before returning.

Removed tab from commented natsJetStreamMaxWaitProcessed option value in config_defaults.go under ers section.

Removed redundant NATS prefix from config.NATSROpts type field names.

Added function in ers/lib_test.go to create a config file in /tmp based on a config string.

Added integration test for ERs NATS.


danbogos:
Thanks!

PR #4110, opened 2023-09-26
Migrate to new jetstream API

ionutboangiu:
Updated ees/ers implementation to use the jetstream package which separates the jetstream context from Core NATS.

Removed the jsOpts fields from the NatsEE struct. We are now using the jetStreamMaxWait option directly through a timeout context.

Added streamName option for NATS reader since it is now required to be specified when creating a consumer (it is not inferred based on subject anymore).

Updated nats ers integration test.


danbogos:
Thanks!

PR #4118, opened 2023-10-03
Update failing versions it test

ionutboangiu:


utils/dataconverter.go
```diff
@@ -586,3 +588,96 @@ func (jsnC JSONConverter) Convert(in any) (any, error) {
...
+		var err error
+		sc.amount, err = strconv.Atoi(paramSlice[3])
+		if err != nil {
+			return nil, err
+		}
+	default:
+		return nil, errors.New("strip converter: invalid number of parameters (should have 3 or 4)")
+	}
+
+	if side := paramSlice[1]; side != MetaLeft &&
```
danbogos, line 625: Please use here utils.IsSliceMember

engine/z_versions_it_test.go
```diff
@@ -27,6 +27,7 @@ import (
 	"os/exec"
```
danbogos, line 27: Please add these tests in a separate PR from your strip so I can accept it.

Thanks!

PR #4140, opened 2023-10-11
Add sanity check to prevent xml reader panic

ionutboangiu:
HierarchyPath parser now returns nil when
the path is empty (instead of a string slice with one EmptyString element).

If the prefix is set to true, when calling the AsString method on a nil HierarchyPath, only the separator will be returned. This avoids a nil expr error coming from the xmlquery library.

Use the Query and QueryAll functions from the xmlquery package to be able to handle the errors ourselves and avoid panics.

Added an integration tes for the special case where the xml_root_path field is left empty. Before the change it used to trim the root element from the path slice when attempting to retrieve a the relative path slice.


danbogos:
Please correct config_defaults and do not modify the way we are parsing it.

Thanks!

PR #4147, opened 2023-10-13
Add sanity check to prevent xml reader panic

ionutboangiu:
HierarchyPath parser now returns nil when
the path is empty (instead of a string slice with one EmptyString element).

If the prefix is set to true, when calling the AsString method on a nil HierarchyPath, only the separator will be returned. This avoids a nil expr error coming from the xmlquery library.

Use the Query and QueryAll functions from the xmlquery package to be able to handle the errors ourselves and avoid panics.

Remove config default value for xml_root_path. The field will remain commented in config_defaults for reference.

Add tests for HierarchyPath.AsString function.

Add comments for XmlProvider and xml_root_path opt.


danbogos:
Thanks!

Issue #4146, opened 2023-10-12
[*ERS] AMQP concurrent_requests basic.ack exception


Integration-IT:
Hi Teams,

Insprired from [AMQPER / amqp_it_test.go](https://github.com/cgrates/cgrates/blob/405f29ca0064254c2a16b69d75d586263d6c07aa/ers/amqp_it_test.go#L36) to configure the consumer, I have unexpectedly closed TCP connection between MQserver and Cgrates.

[*ERS] *amqp_json_map TO [*ESS] *els : ERROR.
[*ERS] *amqp_json_map queue A TO [*ESS] *amqp_json_map queue B : ERROR.

The context is to consume the CDR in queue A with the *ERS, and use the *ESS to push the valuable CDR as *rated mode with cost details to queue B for a third party.
OR
Consume the CDR in queue A with the *ERS, and use the *ESS to push the valuable CDR with cost details to ELK for a third party analysis.

Both contexts are working very well until **concurrent_requests** is set to a very **low value**.
Actualy not possible to replicate the amqp_it_test.go configuration, got closed TCP connection after few seconds from MQServer.

The main vector to replicate is to play with the "concurrent_requests" parameter and keep all others according to the amqp_it_test.go.

---

`CGRateS@v0.11.0~dev`

---
test:
-  ers "concurrent_requests" : 1 --> STABLE
-  ers "concurrent_requests" : 2 --> operation basic.ack caused a channel exception precondition_failed: unknown delivery tag 1
-  ers "concurrent_requests" : 10 --> operation basic.ack caused a channel exception precondition_failed: unknown delivery tag 1, operation basic.ack caused a channel exception precondition_failed: unknown delivery tag 5, operation basic.ack caused a channel exception precondition_failed: unknown delivery tag 3.
-  ers "concurrent_requests" : 100 --> operation basic.ack caused a channel exception precondition_failed: unknown delivery tag 10, operation basic.ack caused a channel exception precondition_failed: unknown delivery tag 19

Integration-IT:
[Delivery Identifiers: Delivery Tags](https://www.rabbitmq.com/confirms.html#consumer-acks-delivery-tags)
[...]

PR #4148, opened 2023-10-16
Update AMQP message processing loop

ionutboangiu:
Previously, msg.Ack(true) was used, which is mostly used for batch processing. It mistakenly acknowledged all previously unacknowledged messages, causing errors from the AMQP server. Now, messages are acknowledged individually after each one is processed.

Messages that ERs failed to process are now rejected and requeued for future processing attempts.

The reader is now closed immediately if the message delivery channel closes. Therefore, it prevents an endless loop by avoiding continuous consumption from empty or closed channels.

Addresses: #4146


danbogos:
Thanks!

PR #4155, opened 2023-10-17
Update kafka ansible role and fix kafka it test

ionutboangiu:
- use kraft instead of zookeeper
- add handlers in case of cfg changes
- create a separate user for the kafka service
- bump kafka version
- make the role more configurable


danbogos:
Thanks!

PR #4178, opened 2023-11-01
Remove unused DryRun field from CallDescriptor

### LINKED ISSUE #4163: Balance blocker is not working properly
I have a problem in engine/account.go
I was looking at codes to see how Blocker works. but in this code while dryrun is disabled, blocker wont work:
```go
// check for blocker
				if dryRun && balance.Blocker {
					return // don't go to next balances
				}
```
Is it right? I think you have to remove dryRun filter or set it as !dryRun

ionutboangiu:
dryRun vars will also not be considered anymore in the balance blocker condition.

Fixes #4163

### DIFF
--- a/engine/account.go
+++ b/engine/account.go
@@ -419,7 +419,7 @@ func (acc *Account) debitCreditBalance(cd *CallDescriptor, count bool, dryRun bo
 					}
 				}
 				// check for blocker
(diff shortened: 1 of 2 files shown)


danbogos:
Thanks!

PR #4198, opened 2023-11-15
Add default evencost  when it's nil for  ees

gezimbll:
it fixes the panic on scenario  when  we define  a template containing *ec fields and exported cdr doesn't contain a costdetails field.


danbogos:
Thanks!

PR #4199, opened 2023-11-15
Update/remove dependencies

ionutboangiu:
Remove cenkalti/rpc2 dependency and update client constructor. 
Commented tests that only start and stop engine.

Fixed compilation errors caused by Azure/go-amqp update.

Remove cgrates/ugocodec dependency since we have everything we need in ugoriji/go/codec and add tests for it.


danbogos:
Thanks!

PR #4200, opened 2023-11-15
Add duration format data converter

arberkatellari:


utils/dataconverter.go
```diff
@@ -274,6 +276,24 @@ func (mS *DurationConverter) Convert(in any) (
 	return IfaceAsDuration(in)
 }
 
+func NewDurationFormatConverter(params string) (hdlr DataConverter, err error) {
+	return new(DurationFormatConverter), nil
+}
+
+// DurationFormatConverter formats duration to hh:mm:ss
```
danbogos, line 283: I think a better explanation should be: formats duration in the same way Time is formatted as string.

danbogos:
Thanks!

danbogos:
Thanks!

PR #4208, opened 2023-11-17
Revise buildURL function

ionutboangiu:
The RawQuery field would not be properly populated in the old implementation. Has been fixed by calling the Parse function at the end.

### DIFF
--- a/engine/storage_utils.go
+++ b/engine/storage_utils.go
@@ -20,8 +20,6 @@ package engine
 
 import (
 	"fmt"
-	"net"
-	"net/url"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


engine/storage_utils.go
```diff
@@ -84,22 +81,19 @@ func NewStorDBConn(dbType, host, port, name, user, pass, marshaler string,
 	return
 }
 
-func buildURL(scheme, host, port, db, user, pass string) (*url.URL, error) {
-	u, err := url.Parse("//" + host)
-	if err != nil {
-		return nil, err
+func composeURI(scheme, host, port, db, user, pass string) string {
```
danbogos, line 84: If this is only used for mongo, please rewrite the function to mongoURI().
Also please do not forget to always add comments to the functions.

Thanks!

danbogos:
Thanks!

PR #4215, opened 2023-11-24
Added OneEvent functionality to StatS

gezimbll:


engine/libstats.go
```diff
@@ -234,15 +234,48 @@ func (sq *StatQueue) TenantID() string {
...
 		return
 	}
+
 	if err = sq.remOnQueueLength(); err != nil {
 		return
 	}
 	return sq.addStatEvent(tnt, evID, filterS, evNm)
 }
 
+func (sq *StatQueue) isOneEvent(tnt string, filterS *FilterS, evNm utils.MapStorage) (bool, error) {
```
danbogos, line 251: isOneEvent should only confirm if we are using special processing or not. There should be no other logic within.

engine/statmetrics.go
```diff
@@ -74,6 +74,7 @@ type StatMetric interface {
 	GetStringValue(roundingDecimal int) (val string)
 	GetFloat64Value(roundingDecimal int) (val float64)
 	AddEvent(evID string, ev utils.DataProvider) error
+	OneEvent(ev utils.DataProvider) error
```
danbogos, line 77: Since all the other methods are having also the verb attached, please rename the method to AddOneEvent

engine/statmetrics.go
```diff
@@ -158,6 +166,22 @@ func (asr *StatASR) AddEvent(evID string, ev utils.DataProvider) (err error) {
 	return
 }
 
+func (asr *StatASR) OneEvent(ev utils.DataProvider) (err error) {
+	var answered int
+	if len(asr.Events) != 0 {
```
danbogos, line 171: We are not allowed to overwrite what we have already,better just take the first event from the queue. So also the asr.Events check against > 1

engine/libstats_test.go
```diff
@@ -730,6 +730,10 @@ func (sMM *statMetricMock) AddEvent(evID string, ev utils.DataProvider) error {
 	return nil
 }
 
+func (sMM *statMetricMock) OneEvent(ev utils.DataProvider) error {
+	return nil
+}
+
 func (sMM *statMetricMock) RemEvent(evTenantID string) error {
 	switch sMM.testcase {
 	case "remExpired error":
```
danbogos, line 739: You would need to modify all the previous tests and also have close to 100% in statmetrics_tests.go before we can accept the PR

danbogos (changes requested):

danbogos:
As discussed, will close the PR here so we can handle all in separate branch until the changes will be completed.

Dan

PR #4224, opened 2023-12-12
Register the CDRServer from cdrs instead of the one from services

ionutboangiu:
Also renamed services.CDRServer to CDRService to prevent such mistakes from happening in the future.


danbogos:
Thanks!

PR #4226, opened 2023-12-13
Update eventcost FieldAsInterface methods

ionutboangiu:
Rating map is now accessible from Accounting.

ExtraCharges struct is accessible from Accounting.

RatingUnit fields that did not represent the id of another event cost struct are now retrievable.


danbogos:
Thanks!

PR #4245, opened 2024-01-24
filter and attribute *http type   fix for  encoding dataprovider as json body &&  integration tests 

gezimbll:


danbogos:
Thanks!

PR #4246, opened 2024-01-24
Implement DisconnectSession API for RADIUS Agent

ionutboangiu:
Updated radigo library to latest version.

Updated RadiusAgent to satisfy the birpc client interface.

Added *radDAdiscMsg OrderedNavigableMap type field within AgentRequest. This one is similar to *diamreq, as it is used for building RADIUS server-initiated Disconnect Requests.

radReplyAppendAttributes: refactored to reflect that it can now be used to also append attributes to request packets, not only reply.

Added bidirectional support for session related RadiusAgent methods.

For Dynamic Authorization to be possible, a new field was added within RadiusAgent that holds dicts and secrets only for the clients that support it. They are used to create the DA Client sending Disconnect Requests.

Added a new cache partition to store Access-Request packets with the purpose of using them to build the Disconnect Requests. They are identified by sessionID. It defaults to the value of 'Acct-Session-id'.

Added a predefined '*dmr' template as well as a 'dmr_template' config option within the 'radius_agent' config section. This will map to a custom or to the predefined template and will be used to build the Disconnect Request. By default, it doesn't point to any template (this also means that the Access-Request packets will not be cached).

Another option added to 'radius_agent' is 'client_da_addresses', which lists the RADIUS clients supporting Dynamic Authorization. The key represents the host of the client, while the value represents the address to which we will send the Disconnect Request.


danbogos:
Thanks!

PR #4247, opened 2024-01-25
Update balance factor to depend on category and SMCostRetries backoff increment unit to ms

ionutboangiu:


danbogos:
Thanks, please add an integration test with .csv definition for the Factor.

PR #4252, opened 2024-02-05
Modernize Debian packaging

sebastic:
The current packaging builds a single binary package for all supported distributions, this will stop working when the build system is upgraded from bullseye to bookworm after which the libc6 dependency will not be satisfiable on bullseye and earlier.

To resolve this, the package should be built for each supported distribution using (cowbuilder) chroots. And the stable/v0.10 and master suites in the apt.cgrates.org repo likewise need to become distribution specific. The ansible playbook has been update for the new build procedure.

The changes in this PR also simplify the git commit date and hash inclusion in the CGRateS version, instead of parsing the `git log` output, the values are passed separately nearly identical to the how this is done for the Debian package version.

The `debian/build.sh` script drives the package build, it creates the `.orig.tar.gz` tarball from the git archive and the additional component tarball for the dependencies, which are then used to create a proper source package. Documentation for the Debian package build is included in `debian/README.source`.

My cowbuilder chroots have the Lintian hook configured to run after every package build, various of these issues are also fixed in separate commits.


danbogos:
Thank you!

sebastic:
Do you need separate PRs for the v0.10 & 1.0 branches?

danbogos:
Hi @sebastic 

It would be great if you could do it, otherwise we will have to replicate your model on v0.10 and 1.0. There is a small difference on v0.10 since that one generates packages only on stable release instead of every time when we run it.

Thanks,
DanB

PR #4256, opened 2024-02-08
Ensure event to be exported is cloned before being sent to attributes + tests

ionutboangiu:


danbogos:
This one is not needed anymore, as discussed.

PR #4257, opened 2024-02-08
Ensure export event is being cloned if processed by attributes

ionutboangiu:


ees/ees.go
```diff
@@ -167,6 +172,9 @@ func (eeS *EventExporterS) V1ProcessEvent(ctx *context.Context, cgrEv *engine.CG
 				return
 			}
 		}
+		if exportEvent == nil {
```
danbogos, line 175: I do not get when will the exportEvent be nil here? If there is no logical reason for it, please let it panic instead of covering it.

danbogos:
Thanks!

ees/ees.go
ionutboangiu: exportEvent will be nil if attributes were not found, in which case we need to overwrite it with the original event. Maybe you can suggest an alternative method for verifying whether AttributeS modified the event or not.

PR #4258, opened 2024-02-09
added *http filter and attribute  type

gezimbll:


danbogos:
Thanks! Please also add integration tests for it.

PR #4259, opened 2024-02-09
Support *any category in RatingPlans retrieval

ionutboangiu:


danbogos:
Since this approach will slow down the queries (2 extra redis queries, 20ms aprox), will discard it in favor of attributes mangling.

PR #4267, opened 2024-02-19
Ensure CostDetails is of type *EventCost when parsing dynamicDP

ionutboangiu:


danbogos:
We decided that we will go with UnmarshalJSON instead of this.

PR #4268, opened 2024-02-21
Modernize Debian packaging (v0.10)

sebastic:
Follow up to #4252, this PR makes the same changes for the v0.10 branch.


ionutboangiu (approved):

data/ansible/deb_packages/main.yaml
```diff
@@ -183,40 +251,73 @@
...
+        path: /var/packages/deb/{{ cgrates_branch }}/{{ item.codename }}/cgrates_current_amd64.deb
         state: absent
-      when: hostvars['apt']['gitTagVar'].stdout_lines|length > 0
+      with_items: "{{ distributions }}"
+      when: cgrates_branch == "master" or hostvars['apt']['gitTagVar'].stdout_lines|length > 0
+
+    - name: Move the new .orig.tar.gz to /var/packages/deb
+      become: yes
+      shell: "mv /tmp/cgrates*.orig*.tar.* /var/packages/deb/{{ cgrates_branch }}/"
+      when: cgrates_branch == "master" or hostvars['apt']['gitTagVar'].stdout_lines|length > 0
```
ionutboangiu, line 274: All tasks that have this condition fail on v0.10 if we are not on a tagged release. The script works perfectly when changing "master" to "v0.10".

sebastic: I thought that was the intention, to publish all package builds of the master branch and only releases for v0.10.

Why have the condition if you want to publish all package builds?

ionutboangiu: As far as I know, it's intended to be used only in the `Include the packages with reprepro` task and for the `current` symlinks on the pkg server, allowing others to test fixes to the current release branch. 

On http://pkg.cgrates.org/deb/v0.10/, we want to provide access to all the packages that have been built. On apt, we only publish releases, and on pkg, the symlink will point to the last release for easy identification. But we can handle it ourselves from here. What's important is that everything else is working perfectly.

PR #4270, opened 2024-02-21
Implement *alter_sessions action and refactor agents apis

ionutboangiu:


engine/action.go
```diff
@@ -849,6 +850,73 @@ func cgrRPCAction(ub *Account, a *Action, acs Actions, _ *FilterS, extraData any
...
+			Tenant:  params[0],
+			Filters: strings.Split(params[1], "&"),
+			APIOpts: make(map[string]any),
+		},
+		Event: make(map[string]any),
+	}
+
+	// Use default tenant if not specified.
+	if attr.Tenant == "" {
+		attr.Tenant = config.CgrConfig().GeneralCfg().DefaultTenant
```
danbogos, line 895: Here should not need to populate it since defaults are taken in the end API.

danbogos:
Thanks!

PR #4273, opened 2024-02-23
Add tests for *destinations filters and attributes arith operations + configsanity fix

ionutboangiu:


danbogos:
Thanks!

sebastic:
The https://github.com/cgrates/cgrates/pull/4273/commits/42de152ea4386c0092bb6082e24a1adcfa273b77 commit is wrong.

The modernized Debian packaging is not native, so the package versions need a revision.

The syntax of the changelog is also wrong, about which `dpkg-parsechangelog` complains.

That commit needs to be reverted.

sebastic:
@ionutboangiu wrote:
> Could you please specify what exactly was the issue with the changelog? Were the issues related to the way change entries were formatted (starting with - instead of *, the lack of indentation etc.) or is it something else?

The UNRELEASED changelog entries should be removed, they are all superseded by the current package build. The changelog never tracked packaging changes properly, hence the modernized packaging doesn't bother with that either.

The syntax also does not conform to the Debian policy [4.4](https://www.debian.org/doc/debian-policy/ch-source.html#debian-changelog-debian-changelog). It should have been:
```
cgrates (0.11.0~dev-1) UNRELEASED; urgency=medium

  * [FilterS] Renamed rals_conns to apiers_conns
  [...]
  * [ActionsS] Replaced structmatcher with normal filters

 -- DanB <danb@cgrates.org> Wed, 19 Feb 2020 13:25:52 +0200
```
The number of spaces at the start of the line are significant; no spaces before the source package and version, two space indent for changelog entries, and one space indent for the signature. The author between square brackets is also useless because its the only contributor of changelog entries.

The `-1` revision makes the package a non-native package (per Debian policy [5.6.12](https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-version)), which is required because native packages don't support multiple upstream tarballs which the modernized packaging uses for the dependencies.

> How should we go about fixing the current changelog?

Revert 158dce732365f4d0e8d36fae708363eeb914ff35 on master, and revert 48f4ef1f90ab75655b81abc08f86845e33096399 on the v0.10 branch.

This PR nor #4275 provides context or justification for reverting the changelog to the old packaging. If you need the old packaging to support customers on buster or older systems, we'll need to re-instate the old packaging differently to make it co-exist with the modern packaging better.

PR #4279, opened 2024-03-01
Set BalanceFactorID from the event directly, update balance actions, ensure CostDetails maintains type

ionutboangiu:


danbogos:
Thanks!

PR #4289, opened 2024-03-11
Add chan_delimiter option within freeswitch_agent section

ionutboangiu:
Used for 'show_channels' requests and responses.


danbogos:
Thanks!

PR #4302, opened 2024-04-02
Implement RunReader API for ERs

ionutboangiu:


ers/filejson.go
```diff
@@ -65,7 +65,7 @@ type JSONFileER struct {
 	cgrCfg        *config.CGRConfig
 	cfgIdx        int // index of config instance within ERsCfg.Readers
 	fltrS         *engine.FilterS
-	rdrDir        string
+	dir           string
```
danbogos, line 68: Is there a strong reason for renaming?

agents/diam_it_test.go
```diff
@@ -1469,39 +1469,27 @@ func testDiamItEmulateTerminate(t *testing.T) {
...
-	chargerProfile := &v1.ChargerWithAPIOpts{
-		ChargerProfile: &engine.ChargerProfile{
-			Tenant:       "cgrates.com",
-			ID:           "CustomCharger",
-			RunID:        "CustomCharger",
-			AttributeIDs: []string{"*constant:*req.Category:custom_charger"},
-			Weight:       20,
-		},
+	//add a default charger
+	tpDirPath := t.TempDir()
```
danbogos, line 1473: Need to check for an option without temporary files which can create leftovers and also unnecessary increase the file descriptors on the server.

utils/file_it_test.go
```diff
@@ -75,9 +75,9 @@ func testWatchWatcherEvents(t *testing.T) {
 		Op:   fsnotify.Create,
 	}
 	stopWatching := make(chan struct{}, 1)
-	f := func(itmPath, itmID string) error {
+	f := func(itmID string) error {
```
danbogos, line 78: Will this not create extra files in the git monitored repo? Temporary folder was there to prevent that.

agents/diam_it_test.go
ionutboangiu: TempDir method on testing.T is the recommended way to create temporary directories during tests. It is accompanied by the following comment:
> TempDir returns a temporary directory for the test to use. The directory is automatically removed when the test and all its subtests complete. Each subsequent call to t.TempDir returns a unique directory; if the directory creation fails, TempDir terminates the test by calling Fatal.

The above behaves similarly to doing:

```go
if err = os.MkdirAll("/tmp/testTPs/", 0700); err != nil {
		t.Fatal(err)
}
```

and then using either defer or t.Cleanup to ensure the files are deleted at the end of the test

```go
defer func () {
    if err := os.RemoveAll("/tmp/testTPs/"); err != nil {
		t.Error(err)
    }
}()
```

ers/filejson.go
ionutboangiu: Not really, it just felt repetitive to always use rdr.rdrDir when adding the new ERs functionality.

As per https://google.github.io/styleguide/go/guide#naming:

> Names should:
> 1. Not feel [repetitive](https://google.github.io/styleguide/go/decisions#repetition) when they are used

The case I talked about above can be included here (rdr.rdrDir can be considered repetitive)
> 2. Take the context into consideration:

dir is found inside a CSV/FWV/JSONFileER which represent file event readers (also implementing the EventReader interface), so we can infer that dir is referring to the directory handled by the reader.
> 3. Not repeat concepts that are already clear


This change should be addressing all of the above.

danbogos:
Thanks!

PR #4307, opened 2024-04-05
Add possibility to set/overwrite balance Factors through API 

ionutboangiu:
Constructor looks inside the params' balance map for the Factors key.
It expects either a string representing a JSON serialized map or the
map itself.

BalanceFilter Clone function has been updated to set a deep copy of
the original Factors map instead of a shallow one.

BalanceFilter getter function for Factors now returns nil instead of
an empty map. It's slightly more memory efficient and assignment to
this map will not be attempted, so it's panic proof.

BalanceFilter.ModifyBalance now updates Factors only if the key is
found in the request params' Balance map. Setting Factors to null is also
possible as long as the Factors key exists and is set to null.

Note: only *set_balance can overwrite the Factors map, all the others
can only set it if the balance does exist prior to sending the request.

Update balance integration tests.


danbogos:
Thanks!

PR #4314, opened 2024-04-10
Ensure EC FieldAsInterface method works correctly

ionutboangiu:
- trim fldPath Increments before retrieving the field
- Rating map can be accessed from Accounting
- ExtraCharges can be accessed from Accounting
- RatingUnit fields that did not represent the id of another EC struct are now retrievable
- add unit tests


danbogos:
Thanks!

PR #4315, opened 2024-04-10
Set kafka MaxAttempts to 1

ionutboangiu:
This leaves it to the ExportWithAttempts function to handle the connect attempts.


danbogos:
Thanks!

PR #4323, opened 2024-04-17
Don't trim single digit values when parsing cron

ionutboangiu:
Fixes an issue where 0 values would become empty inside the cron expressions.

Added unit tests for the edge cases and grouped them together with the previous ones under the same table test.

When creating the StartTime field, assign time.Now() to a centralised variable and reuse it instead of calling time.Now() repeatedly.


danbogos:
Thanks!

PR #4324, opened 2024-04-21
Revise docker ansible script

ionutboangiu:
- use include/import_tasks instead of the deprecated include.

Used include_tasks for nginx tasks because they might cause problems due to depending on the nginxConfig variable inside the conditional statements.
Used import_tasks for docker tasks since they seem like they can be safely pre-processed.

- remove task installing docker module for python


danbogos:
Thanks!

PR #4329, opened 2024-05-03
revise threshold snooze time to update only after sleep interval

gezimbll:


danbogos:
Please add Snooze after executing actions since in case when an action takes longer to execute as sleep is configured, sleep will have no efect.

PR #4358, opened 2024-05-29
Don't recursively change ownership of package directories.

sebastic:
The Debian package maintainer script should not recursively change the ownership on the package directories.

The /var/spool/cgrates on our systems is very large, which causes this operation to take a very long time, making package upgrades a PITA because the cgrates service becomes unresponsive to the cluster resource manager.

The ownership of /usr/share/cgrates shouldn't be changed at all, because /usr can be mounted read-only, cgrates should not be writing to that directory tree.

Instead of using `chown -R` the directories created by the package are acted on individually.

Note: This change need to be applied to the 1.0 and v0.10 branches too.

### DIFF
--- a/debian/lintian-overrides
+++ b/debian/lintian-overrides
@@ -10,9 +10,6 @@ package-contains-documentation-outside-usr-share-doc [usr/share/cgrates/*]
 # Only systemd is supported
 package-supports-alternative-init-but-no-init.d-script [lib/systemd/system/cgrates.service]
 
-# No good alternative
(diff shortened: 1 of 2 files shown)


ionutboangiu (approved):

danbogos:
Thanks!

sebastic:
Do you need separate PRs for the other branches?

danbogos:
We have opened an internal issue to fix them. If you need them faster, please let us know.

Thanks!

Issue #4357, opened 2024-05-28
The same filter update by SetFilter method enables full cache load


vicshulha:
Hi All,

Upon recommendation from Ionuț at https://groups.google.com/g/cgrates I am creating the issue report here.

We face a new issue loading the full Redis cache during SetFilter method usage. It started to happen after CGRates version update from CGR...@v0.10.2~dev-20200831063159-3289a59a41c3 to the latest one (CGR...@v0.10.3).

The issue symptoms and negative effects:

Redis cache load generates huge network traffic (1.5G data sent by 60mbit/s for 2 min.) between CGrates and Redis nodes. In turn, this behavior leads to CGRates performance degradation which fails with “ATTRIBUTES_ERROR:context deadline exceeded” error and does not allow processing a part of new requests for some time.

Steps to reproduce (STR) the issue:

1) Create attribute profile (SetAttributeProfile.txt);
2) Set a filter to add 1k numbers (SetFilter_1.txt);
3) Set the same filter with 1 number to delete the rest of 999 ones (SetFilter_2.txt);

After the 3rd step, the issue symptoms become observable namely huge network traffic starts, and CGRates is loaded (see CPU statistics). If to add 1k numbers again (step 2) after the STR, no cache load process has started but only reappears after step 3.

Please find the files used during STR uploaded to GDrive (https://drive.google.com/drive/folders/1cvlzfYa0rKQQM3cxwvojiKM0Py0E4FX0?usp=drive_link):

- ngrep log (6379,2080,2012 ports);
- requests used in STR;
- cgrates.json, cgrates.cfg (kamailio);
- network, CPU statistics screen for 2 min.

I would appreciate an idea of what CGRates change led to this behavior and how it can be fixed.

ionutboangiu:
Hi,

I tried to follow the steps you provided but couldn't reproduce the issue. 

1.  The Set* API requests you sent reload the cache by default unless configured otherwise. This is determined by the "default_caching" setting in the JSON configuration file (can be overwritten by the cache specific request parameter). Since this setting wasn't in your configuration, the default *reload behavior happened.
    
2.  When I followed your steps, the process was quick for me because I don't have much cached data.
    
3.  It's normal to see higher network traffic during the setup phase as the cache is being filled/reloaded and the indexes are being built (as long as it doesn't break the system). This is to make sure outdated data is not being used.
    
Could you please tell me:

-   How long did you experience the high network traffic?
-   Did you have any performance issues after the cache reload finished?

[...]

PR #4362, opened 2024-06-05
Optimize removeFilterIndexesForFilter function + tests

ionutboangiu:
Previously made two trips (get and set) for each group of items from an index key being removed. Now, we fetch indexes once at the beginning and store the updated indexes once at the end.

Related to #4357


danbogos:
Thanks!

PR #4364, opened 2024-06-07
added aliases for extra_fields

gezimbll:


engine/fscdr.go
```diff
@@ -77,6 +77,7 @@ func (fsCdr FSCdr) getCGRID() string {
 
 func (fsCdr FSCdr) getExtraFields() map[string]string {
 	extraFields := make(map[string]string, len(fsCdr.cgrCfg.CdrsCfg().ExtraFields))
+	extraFieldsAliases := fsCdr.cgrCfg.CdrsCfg().ExtraFieldsAliases
```
danbogos, line 80: Here is no need of extra copy, you can directly use it from the config.

engine/fscdr.go
```diff
@@ -86,9 +87,12 @@ func (fsCdr FSCdr) getExtraFields() map[string]string {
 			origFieldVal = fsCdr.searchExtraField(field.Id, fsCdr.body)
 		}
 		if parsed, err := field.Parse(origFieldVal); err == nil {
+			if alias, has := extraFieldsAliases[field.Id]; has && alias != utils.EmptyString {
+				extraFields[alias] = parsed
```
danbogos, line 91: Please unite both into one, no need of separate continue inside since it is a bit confusing. I would go like:
fldName = field.ID
if hasAlias -> {fldName = alias}
extraFields[fldName] = parsed

engine/fscdr.go
```diff
@@ -77,21 +77,20 @@ func (fsCdr FSCdr) getCGRID() string {
...
 	extraFields := make(map[string]string, len(fsCdr.cgrCfg.CdrsCfg().ExtraFields))
-	extraFieldsAliases := fsCdr.cgrCfg.CdrsCfg().ExtraFieldsAliases
 	for _, field := range fsCdr.cgrCfg.CdrsCfg().ExtraFields {
-		origFieldVal, foundInVars := fsCdr.vars[field.Id]
+		fldName := field.Id
+		origFieldVal, foundInVars := fsCdr.vars[fldName]
 		if strings.HasPrefix(field.Id, utils.STATIC_VALUE_PREFIX) { // Support for static values injected in the CDRS. it will show up as {^value:value}
 			foundInVars = true
 		}
 		if !foundInVars {
```
danbogos, line 86: Why not using here alias as well?

danbogos:
Thanks!

PR #4366, opened 2024-06-07
Revise test for ReplayFailedPosts API

ionutboangiu:
- shortened the failed post cache ttl to 1ms to simulate with more than one events file
- use the default FailedPath (which is the same as SourcePath) to check whether writing events that failed to be replayed to the same path would cause an infinite loop.


danbogos:
Thanks!

PR #4368, opened 2024-06-10
Implemented service,config and api for Sars

gezimbll:


apier/v1/sars.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"github.com/cgrates/birpc/context"
+	"github.com/cgrates/cgrates/utils"
+)
+
+func NewSArSv1() *SaRSv1 {
+	return &SaRSv1{}
+}
+
+type SaRSv1 struct {
```
danbogos, line 30: Should be named SarS instead

danbogos:
Thanks!

PR #4374, opened 2024-06-11
Add engine flag to print the config object in JSON format

ionutboangiu:
Removed the redundant check_config engine flag. Config sanity was checked by the NewCGRConfigFromPath function.


cmd/cgr-engine/cgr-engine.go
```diff
@@ -439,6 +433,11 @@ func main() {
 	}
 	utils.Logger.SetLogLevel(lgLevel)
 
+	if *printConfig {
+		cfgJSON := utils.ToIJSON(cfg.AsMapInterface(cfg.GeneralCfg().RSRSep))
+		utils.Logger.Info(fmt.Sprintf("Configuration loaded from %s:\n%s", *cfgPath, cfgJSON))
```
danbogos, line 438: Please add here the path between <>, something like: Configuration loaded from <%s> ...

danbogos:
Thanks!

PR #4377, opened 2024-06-13
Add reference value functionality to *transfer_balance action

ionutboangiu:
The *transfer_balance action can now use a reference value to ensure the destination balance reaches a specified amount. If the destination balance exceeds the reference value, the excess is transferred back to the source balance. If the destination balance is below the reference value, the required amount is transferred from the source balance to the destination balance to reach the specified reference value. An error is returned if the transfer cannot achieve the specified reference value.

Used by specifying DestinationReferenceValue inside ExtraParameters.

Other *transfer_balance changes:
- used json tags when unmarshaling ExtraParameters in order to be able to shorten the names of the fields
- lock the destination account only if it's different from the source account. It is still passed to the Guard function but without a lock key and with 0 timeout.
- if the transfer happens within the same account, update the account and execute its ActionTriggers only once.
- moved transfer units validation after retrieving/creating the destination balance

*cdrlog action has been updated to create cdrs for reference *transfer_balance actions, although improvements are needed and the functionality is not completely tested.

APIerSv1.TransferBalance has been updated to take into account the ReferenceValue parameter.

Added new *transfer_balance action unit tests to account for the new changes.

Added integration tests (incomplete for now, but functionality has been tested manually).

### DIFF
--- a/apier/v1/accounts.go
+++ b/apier/v1/accounts.go
@@ -714,7 +714,7 @@ func (apierSv1 *APIerSv1) RemoveBalances(ctx *context.Context, attr *utils.AttrS
 }
 
 // TransferBalance initiates a balance transfer between accounts, immediately executing the configured actions.
-func (apierSv1 *APIerSv1) TransferBalance(ctx *context.Context, attr utils.AttrTransferBalance, reply *string) (err error) {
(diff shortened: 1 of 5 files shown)


danbogos:
Thanks!

PR #4382, opened 2024-06-17
Ensure config path stands out when using print_config flag

ionutboangiu:
Done by using the %q fmt verb instead of %s, which wraps the path within quotes.


danbogos:
Thanks!

PR #4384, opened 2024-06-18
Update ActionTrigger integration test scenario

ionutboangiu:
Instead of using 2 balances and maintaining them using *transfer_balance, use only 1 balance.


danbogos:
Thanks!

PR #4390, opened 2024-06-21
Add support for Git 2.45+ iso-strict date format

ionutboangiu:
Git 2.45+ introduced a backward incompatible change in the iso-strict date format, showing time in the Zulu timezone with Z suffix instead of +00:00. This commit adds parsing for the new date format before falling back to the old format.

Ignore errors for old git versions where %cI is not defined.

Revise GetCGRVersion error messages.
Revise GetCGRVersion unit test.


danbogos:
Thanks!

PR #4400, opened 2024-07-02
Add redisPoolPipelineWindow configuration field for datadb

ionutboangiu:
Sets the duration after which internal pipelines are flushed. 0 disabled implicit pipelining.


danbogos:
Thanks!

PR #4404, opened 2024-07-04
Add precache support for exporters

ionutboangiu:
Additional changes:
- removed unnecessary locking when initializing the exporter cache map, as the service itself is responsible for locking.
- separated setupCache method into ClearExporterCache and SetupExporterCache methods.
- removed idle ListenAndServe function that was only waiting for the stopChan to close. The reload case was unreachable due to the reload channel being created in Start instead of being passed down.
- removed Shutdown method on EventExporterS and replaced it with the exported ClearExporterCache method as it provided the same functionality.

### DIFF
--- a/ees/ees.go
+++ b/ees/ees.go
@@ -38,17 +38,18 @@ func onCacheEvicted(_ string, value any) {
 	ee.Close()
 }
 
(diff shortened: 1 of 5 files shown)


danbogos:
Thanks!

PR #4405, opened 2024-07-04
Revise debitCreditBalance

arberkatellari:
Revised to return maxSessionTime instead of error when balance block is hit


danbogos:
Thanks!

PR #4406, opened 2024-07-04
Sync (and update) go version 

ionutboangiu:
go.mod had a later version than the one used inside the
role, which caused the deb packaging to fail.


danbogos:
Thanks!

PR #4407, opened 2024-07-04
Add redisPoolPipelineLimit configuration field for datadb

ionutboangiu:
Sets the maximum number of commands that can be pipelined before flushing. 0 means no limit and pipelines will only be limited by the time window.


danbogos:
Thanks!

PR #4410, opened 2024-07-08
Ensure refund flag is not ignored when rerate is true

ionutboangiu:
By default setting rerate to true also sets refund to true, but flags should take precedence over defaults.

If rerate is true and refund is false, remove any previous CostDetails from event to force rerate.

Centralize the parsing of processing flags.


danbogos:
Thanks!

PR #4414, opened 2024-07-10
Remove CGRConfig and error global variables used in tests

ionutboangiu:
Removed many leftover error checks for NewDefaultCGRConfig,
which does not return an error anymore.


danbogos:
Thanks!

PR #4416, opened 2024-07-12
Add set_versions flag to cgr-engine

ionutboangiu:
If set to true, it sets/overwrites all versions to the current ones.


danbogos:
Thanks!

PR #4421, opened 2024-07-18
Refactor profiling implementations and flag handling

ionutboangiu:


cmd/cgr-engine/cgr-engine.go
```diff
@@ -330,9 +330,7 @@ func runPreload(loader *services.LoaderService, internalLoaderSChan chan birpc.C
 }
 
 func main() {
-	if err := cgrEngineFlags.Parse(os.Args[1:]); err != nil {
-		log.Fatalf("<%s> error received: <%s>, exiting!", utils.InitS, err.Error())
-	}
+	cgrEngineFlags.Parse(os.Args[1:])
```
danbogos, line 333: Can the flags not return error here?

PR #4472, opened 2024-08-30
disable automatic removal of expired balances in test

ionutboangiu:
instead of setting the expiry time sometime in the future and then waiting for it to expire, now the balances are set directly with expired balances


danbogos:
Thanks!

PR #4478, opened 2024-09-06
remove unused call & redundant ping methods

ionutboangiu:
- Ping methods defined by engine.NewService for all services
- Call methods were previously replaced by the centralized call method (on *birpc.Service) for all components
- All services now register V1/V2 objects from apier package for consistency


danbogos:
Thanks!

PR #4482, opened 2024-09-06
improve CoreSv1.Status API with more detailed metrics

ionutboangiu:
- add new metrics (off by default)
- add Debug boolean option for full metric details
- update default metrics shown
- remove utils.GetStartTime(), now use process metrics


danbogos:
Thanks!

PR #4493, opened 2024-09-17
Authorize pseudoprepaid ProcessMessage requests

ionutboangiu:
Previously, only *prepaid requests were authorized, for every other request type, the default usage was used.


danbogos:
Thanks!

PR #4501, opened 2024-09-19
Fix flaky nats integration test

ionutboangiu:
Was failing on the Jetstream JWTAuth case with error:

nats: API error: code=500 err_code=10047 description=insufficient storage resources available

The issue was with the previously generated credentials. Generated them again and added the program used as a comment at the end of the file for future reference"


danbogos:
Thanks!

PR #4512, opened 2024-09-26
added it tests and new APIs for TrendS

gezimbll:


danbogos:
Thanks!

engine/trends.go
```diff
@@ -39,8 +39,9 @@ func NewTrendS(dm *DataManager,
 		connMgr:     connMgr,
 		filterS:     filterS,
 		cgrcfg:      cgrcfg,
+		crn:         cron.New(),
 		loopStopped: make(chan struct{}),
-		crnTQsMux:   new(sync.RWMutex),
+		crnTQsMux:   &sync.RWMutex{},
```
danbogos, line 44: Please keep new to init here, as originally written. Rule out of thumb in our code is to use new for instantiating an empty object and struct literal whenever we are populating also  parameters inside.

PR #4518, opened 2024-10-03
Improve engine setup helpers

ionutboangiu:
- added hook support (executed after parsing config but before starting engine)
- made db resets configurable
- merged config parsing helper with the main Setup function
- renamed TestEnvironment.Setup -> TestEngine.Run as it represents the setup for a single cgr-engine instance
- removed engineDelay parameter. Added helper to wait up to 200ms for the APIerSv1 service to be up and running to prevent 'can't find service' errors.
- replaced t.Log with t.Error for engine process kill error
- improved option comments


danbogos:
Thanks!

PR #4527, opened 2024-10-09
added integration tests for trend event to thresholds,EEs && fixes

gezimbll:


engine/libtrends.go
```diff
@@ -120,12 +122,48 @@ func (t *Trend) Clone() (tC *Trend) {
 	return
 }
 
+func (t *Trend) compress(ms Marshaler) (err error) {
+	maxRL := config.CgrConfig().TrendSCfg().StoreUncompressedLimit
+	if maxRL > 0 && maxRL > len(t.RunTimes) {
```
danbogos, line 127: Here only one comparison should be enough since len(t.RunTimes) can never be lower than 0.

danbogos:
Thanks!

PR #4533, opened 2024-10-14
moved trend compressing in datamanager

gezimbll:


engine/datamanager.go
```diff
@@ -1345,6 +1338,9 @@ func (dm *DataManager) GetTrend(tenant, id string,
 			return
 		}
 	}
+	if err := tr.uncompress(dm.ms); err != nil {
```
danbogos, line 1341: Here you do not need to redefine err since it is already defined in the function's signature.

engine/libtrends.go
```diff
@@ -137,8 +136,7 @@ func (t *Trend) compress(ms Marshaler) (err error) {
 }
 
 func (t *Trend) uncompress(ms Marshaler) (err error) {
-	maxL := config.CgrConfig().TrendSCfg().StoreUncompressedLimit
-	if t == nil || (maxL > 0 && t.RunTimes != nil) {
+	if t == nil || t.RunTimes != nil {
```
danbogos, line 139: As discussed, use here len(CompressedMetrics) instead.

danbogos:
Thanks!

PR #4541, opened 2024-10-17
Adding rankings to config_defaults 

armirveliaj:


config/config_defaults.go
```diff
@@ -1187,6 +1187,11 @@ const CGRATES_CFG_JSON = `
 "rankings":{								// RankingS config
```
danbogos, line 1187: Please keep the right order in both config_defaults as well as in the config structures. The order should follow the one in trends.

danbogos:
Thanks!

PR #4543, opened 2024-10-18
Revise TrendS && RankingS config

armirveliaj:
-  Add store_uncompressed_limit option to TrendS
-  Revise order in config_defaults and struts for TrendS && RankingS


danbogos:
Thanks!

PR #4554, opened 2024-10-30
rpc_conns: add docs section about bidirectional conns

ionutboangiu:
Also included details about the *http_jsonrpc codec


danbogos:
Thanks!

PR #4560, opened 2024-11-04
Extract cores.Server into new commonlisteners package

ionutboangiu:
Functionality is the same.
Server has been renamed to CommonListenerS.


danbogos:
Thanks!

PR #4564, opened 2024-11-07
Ensure services wait for AnalyzerS to be initiated + refactors

ionutboangiu:


danbogos:
Please refactor as discussed. There should be no "WaitForX" function but simple read out of the service channel.

PR #4573, opened 2024-11-14
Add WHERE statement availability and non-delete option to ERS SQL reader

arberkatellari:


ers/sql.go
```diff
@@ -121,11 +122,185 @@ func (rdr *SQLEventReader) Serve() (err error) {
...
+			rdr.rdrErr <- err
+			return
+		} else {
+			filters = append(filters, result)
+			if !strings.Contains(fltr, utils.DynamicDataPrefix+utils.MetaReq) {
+				renewedFltrs = append(renewedFltrs, fltr)
+			}
+		}
+	}
+	rdr.Config().Filters = renewedFltrs // remove filters containing *req
```
danbogos, line 277: Please find another solution here since we cannot overwrite config.

ers/sql.go
```diff
@@ -121,11 +122,185 @@ func (rdr *SQLEventReader) Serve() (err error) {
...
+			}
+			if elem == utils.EmptyString {
+				singleCond = fmt.Sprintf(" %s = '%s'", field, value)
+			} else {
+				singleCond = fmt.Sprintf(" JSON_VALUE(%s, '$.%s') = '%s'", elem, field, value)
+			}
+		case utils.MetaLessThan, utils.MetaLessOrEqual, utils.MetaGreaterThan, utils.MetaGreaterOrEqual:
+			if ruleType == utils.MetaGreaterOrEqual {
+				if elem == utils.EmptyString {
+					singleCond = fmt.Sprintf(" %s >= %s", field, value)
```
danbogos, line 189: Please do not start with empty string, put instead space when building the query.

ers/sql.go
```diff
@@ -121,11 +122,185 @@ func (rdr *SQLEventReader) Serve() (err error) {
...
+	var filters []*engine.Filter
+	var whereQueries []string
+	var renewedFltrs []string
+	for _, fltr := range rdr.Config().Filters {
+		if result, err := rdr.dm.GetFilter(config.CgrConfig().GeneralCfg().DefaultTenant, fltr, true, false, utils.NonTransactional); err != nil {
+			rdr.rdrErr <- err
+			return
+		} else {
+			filters = append(filters, result)
+			if !strings.Contains(fltr, utils.DynamicDataPrefix+utils.MetaReq) {
```
danbogos, line 272: You should only look for prefix of the element, not contains. Plus you are only checking the inline filters in this way.

ers/sql.go
```diff
@@ -121,11 +122,185 @@ func (rdr *SQLEventReader) Serve() (err error) {
...
+			}
+		}
+	}
+	rdr.Config().Filters = renewedFltrs // remove filters containing *req
+	for _, filter := range filters {
+		for _, rule := range filter.Rules {
+			var elem, field string
+			switch {
+			case strings.HasPrefix(rule.Element, utils.DynamicDataPrefix+utils.MetaReq+utils.NestingSep):
+				field = strings.TrimPrefix(rule.Element, utils.DynamicDataPrefix+utils.MetaReq+utils.NestingSep)
```
danbogos, line 283: When splitting, use constants, not strings here.

danbogos (changes requested):

ers/sql.go
```diff
@@ -121,6 +124,141 @@ func (rdr *SQLEventReader) Serve() (err error) {
 	return
 }
 
+// Creates mysql conditions used in WHERE statement out of filters
+func valueQry(ruleType, elem, field string, values []string, not bool) (conditions []string) {
```
danbogos, line 128: I would vote to move this function in utils so we can use it in the future for other cases (ie: EEs).

ers/sql.go
```diff
@@ -134,8 +272,43 @@ func (rdr *SQLEventReader) readLoop(db *gorm.DB, sqlDB io.Closer) {
...
+			filters = append(filters, resultFltr)
+		}
+	}
+	for _, filter := range filters {
+		var addFltrCount int
+		for _, rule := range filter.Rules {
+			var elem, field string
+			switch {
+			case strings.HasPrefix(rule.Element, utils.DynamicDataPrefix+utils.MetaReq+utils.NestingSep):
+				field = strings.TrimPrefix(rule.Element, utils.DynamicDataPrefix+utils.MetaReq+utils.NestingSep)
```
danbogos, line 290: Somehow this part (with Trim) asks for a better approach, maybe we can find it.

ers/sql.go
```diff
@@ -169,13 +342,14 @@ func (rdr *SQLEventReader) readLoop(db *gorm.DB, sqlDB io.Closer) {
...
 			}
+			// copy row values to their respective column
 			if err = rows.Scan(columnPointers...); err != nil {
 				rdr.rdrErr <- err
 				rows.Close()
 				return
 			}
 			msg := make(map[string]any)
-			fltr := make(map[string]string)
+			fltr := make(map[string]any)
```
danbogos, line 352: If possible, please redefine the filter related variables with more suggestive names.

ers/sql.go
```diff
@@ -184,23 +358,40 @@ func (rdr *SQLEventReader) readLoop(db *gorm.DB, sqlDB io.Closer) {
...
 				}
 			}
-			if err = db.Table(rdr.tableName).Delete(nil, fltr).Error; err != nil { // to ensure we don't read it again
-				utils.Logger.Warning(
-					fmt.Sprintf("<%s> deleting message %s error: %s",
-						utils.ERs, utils.ToJSON(msg), err.Error()))
-				rdr.rdrErr <- err
-				rows.Close()
-				return
+			if rdr.Config().Opts.SQL.DeleteIndexedFields != nil {
```
danbogos, line 375: Here I would mix the if conditions in different way:
1. Only process the delete part of processedPath == *metaDelete.
2. Check first indexed fields.
3. If filterMp is still empty, add the fields out of colums.

danbogos (changes requested):

engine/filters.go
```diff
@@ -790,3 +790,13 @@ func (fltr *FilterRule) passHttp(dDP utils.DataProvider) (bool, error) {
 	return filterHTTP(fltr.Type, dDP, fltr.Element, strVal)
 
 }
+
+// Will split the elemnt, starting after its prefix, on each NestingSep
+func (fltr *FilterRule) SplitElementsField() (string, string) {
```
danbogos, line 795: For core functions like this one, please add few tests to document it.

danbogos:
Thanks!

utils/dbutils.go
```diff
@@ -0,0 +1,156 @@
...
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>
+*/
+
+package utils
+
+import "fmt"
+
+// Creates mysql conditions used in WHERE statement out of filters
+func FilterToSQLQuery(ruleType, beforeSep, afterSep string, values []string, not bool) (conditions []string) {
```
danbogos, line 24: Should this one not be a method on a filter maybe? In this way we are sure what we are doing.
Even if not, please check again beforeSep and afterSep variables since they do not make much sense.

PR #4580, opened 2024-11-26
Revise go role cache clean tasks

ionutboangiu:
Use go env GOMODCACHE for path resolution
Add become: true to prevent permission issues
Check stdout instead of stderr for go clean -x --cache


danbogos:
Thanks!

PR #4581, opened 2024-11-26
revise tests for trends storeinterval && file readers

gezimbll:


engine/datamanager.go
```diff
@@ -1367,7 +1366,7 @@ func (dm *DataManager) SetTrend(tr *Trend) (err error) {
 		return utils.ErrNoDatabaseConn
 	}
 	if dm.dataDB.GetStorageType() != utils.MetaInternal {
-		if err = tr.compress(dm.ms); err != nil {
+		if tr, err = tr.compress(dm.ms); err != nil {
```
danbogos, line 1369: Please create here a new variable instead of overwriting the old one, to be cleaner. Ie: trStored.

danbogos:
Thanks!

PR #4587, opened 2024-12-03
Remove services/cgr-engine.go

ionutboangiu:
Trying to move services init before CGREngine constructor makes it pointless, as it's only used as a dependency storage for services init.


danbogos:
Thanks!

PR #4594, opened 2024-12-06
Remove anzCh in favor of using the service indexer

ionutboangiu:


services/attributes.go
```diff
@@ -98,8 +96,10 @@ func (attrS *AttributeService) Start(ctx *context.Context, _ context.CancelFunc)
 	if datadb, err = attrS.dm.WaitForDM(ctx); err != nil {
 		return
 	}
-	anz := <-attrS.anzChan
-	attrS.anzChan <- anz
+	anz := attrS.serviceIndexer.GetService(utils.AnalyzerS).(*AnalyzerService)
```
danbogos, line 99: Here you should be OK with returning the service, no need of casting into AnalyzerService, no?

danbogos:
Thanks!

PR #4595, opened 2024-12-10
Allow configurable event field updates in ARI sessions

ionutboangiu:


danbogos:
Please add a real-life scenario with real updates/destroy instead of field1/field2 in tests. Thanks!

PR #4610, opened 2024-12-17
Add ability to ERS to update of move ERS SQL events to a new table & add *export flag for ERS readers

arberkatellari:


config/erscfg.go
```diff
@@ -922,10 +922,10 @@ func (er *EventReaderCfg) AsMapInterface(separator string) (initialMP map[string
 		if sqlOpts.TableName != nil {
 			opts[utils.SQLTableNameOpt] = *sqlOpts.TableName
 		}
-		if sqlOpts.DeleteIndexedFields != nil {
-			deleteIndexedFields := make([]string, len(*sqlOpts.DeleteIndexedFields))
-			copy(deleteIndexedFields, *sqlOpts.DeleteIndexedFields)
-			opts[utils.SQLDeleteIndexedFieldsOpt] = deleteIndexedFields
+		if sqlOpts.WhereIndexedFields != nil {
```
danbogos, line 925: Is this as where still needed, or only Delete possible?

ers/nats_it_test.go
```diff
@@ -150,8 +150,8 @@ var natsCfg string = `{
 			"id": "nats_reader",
 			"type": "*nats_json_map",
 			"source_path": "%s",
-			"ees_success_ids": ["nats_processed"],
-			"flags": ["*dryrun"],
+			"ees_success_ids": [],
+			"flags": ["*dryrun", "*export:nats_processed"],
```
danbogos, line 154: This approach does not look like achieving the old behavior with success/fail ids.

data/conf/samples/ers_mysql_update/cgrates.json
```diff
@@ -0,0 +1,84 @@
...
+				"run_delay": "1m",									
+				"source_path": "*mysql://cgrates:CGRateS.org@127.0.0.1:3306",					
+				"opts": {
+					"sqlDBName":"cgrates2",
+					"sqlTableName":"cdrs", 
+				},
+				"start_delay": "500ms", // wait for db to be populated before starting reader 
+				"tenant": "cgrates.org",							
+				"filters": [
+				"*gt:~*req.answer_time:-168h", // dont process cdrs with answer_time older than 7 days ago
```
danbogos, line 56: Small note here, the best would be to indent the filters with a tab or something since they are on the same level.

danbogos:
Thanks!

PR #4624, opened 2025-01-16
added rsparser field in dynamicOpts

gezimbll:
changed Value field to unexported and created a Value() method on each dynOpts type that contains logic to create a rsparser in case the value field a configuration starts with prefix "~" otherwise it will return the specified value


danbogos:
Thanks!

PR #4629, opened 2025-01-21
Adding *sessionSOriginID StringOpts variables

armirveliaj:


danbogos:
Since the suffix "S" is part of the module name, please rename from *sessionSOriginID to *sessionsOriginID. Thanks!

danbogos:
Thanks!

PR #4644, opened 2025-02-04
Revise StateDeps implementation

ionutboangiu:


services/commonlisteners.go
```diff
@@ -100,7 +100,7 @@ func (s *CommonListenerService) Shutdown(registry *servmanager.ServiceRegistry)
 		utils.TrendS,
 	}
 	for _, svcID := range deps {
-		if !servmanager.IsServiceInState(registry.Lookup(svcID), utils.StateServiceUP) {
+		if servmanager.State(registry.Lookup(svcID)) != utils.StateServiceUP {
```
danbogos, line 103: Please rename to ServiceState since on it's own it does not say much.

PR #4650, opened 2025-02-14
Extract rsrparser.go and objectdp.go to utils package

ionutboangiu:
- allows RSRParser to be split gracefully from config (was depending on RSRSep, which was retrieved from the global config variable)
- default separator (;) was the only one used except for splitting inline attribute values (where ANDSep is used)


danbogos:
Thanks!

PR #4681, opened 2025-03-05
pass cfg to DataManager and move rankings/trends to dedicated packages

ionutboangiu:
instead of just CacheCfg. Prevents the need of global config var in engine/datamanager.go


danbogos:
Thanks!

PR #4687, opened 2025-03-12
added option to compress CostDetails before stored in db

gezimbll:


danbogos (changes requested):

engine/cdr.go
```diff
@@ -344,7 +476,26 @@ func NewCDRFromSQL(cdrSQL *CDRsql) (cdr *CDR, err error) {
 			return nil, err
 		}
 	}
-	if cdrSQL.CostDetails != "" {
+	if config.CgrConfig().CdrsCfg().CompressStoredCost && cdrSQL.CostDetails != utils.EmptyString {
```
danbogos, line 479: Please integrate it in a more logical way in the if here, for example:

if cdrSQL.CostDetails != utils.EmptyString {
    if compressed { ...} 
    } else {
    }
}

danbogos:
Thanks!

PR #4697, opened 2025-03-19
add gpg_key_rotation ansible playbook

ionutboangiu:
ensures all repo versions are signed with the current GPG key, without having to build new packages.


ionutboangiu:
@sebastic

sebastic:
The playbook looks to be doing what it needs to do. `reprepro export <dist>` will get the Release.gpg file updated.

The duplication of tasks with the deb_packages playbook is only not ideal.

danbogos:
Thanks!

PR #4702, opened 2025-03-21
add CoreS metrics collection to PrometheusAgent

ionutboangiu:


config/configsanity.go
```diff
@@ -1094,6 +1094,9 @@ func (cfg *CGRConfig) checkConfigSanity() error {
 			return fmt.Errorf("<%s> the CleanupInterval needs to be bigger than 0", utils.AnalyzerS)
 		}
 	}
+	if err := cfg.prometheusAgentCfg.validate(cfg); err != nil {
```
danbogos, line 1097: Please redesign validate since it is taking the main object as argument which is twisted.

danbogos:
Thanks!

PR #4736, opened 2025-04-25
add prometheus_agent docs

ionutboangiu:


docs/prometheus.rst
```diff
@@ -0,0 +1,105 @@
...
+        # HELP go_memstats_alloc_bytes Number of bytes allocated in heap and currently in use.
+        # TYPE go_memstats_alloc_bytes gauge
+        go_memstats_alloc_bytes{node_id="e94160b"} 1.1360808e+07
+
+How It Works
+------------
+
+The PrometheusAgent operates differently than other CGRateS components that use connection failover:
+
+- When multiple connections are configured in stats_conns, the agent sends metrics collection requests to **all** connections, not just the first available one
```
danbogos, line 97: I think here you wanted to say gathers metrics from all stat servers instead of sending them to stats?

docs/prometheus.rst
```diff
@@ -0,0 +1,105 @@
...
+- When multiple connections are configured in cores_conns, the agent attempts to collect metrics from **all** connections, labeling them with their respective node_id
+- The agent processes metrics requests only when Prometheus sends a scrape request to the configured HTTP endpoint
+
+You can view all exported metrics and see what Prometheus would scrape by making a simple curl request to the HTTP endpoint:
+
+.. code-block:: bash
+
+    curl http://localhost:2080/metrics
```
danbogos, line 105: the default url here is too generic and it overlaps with the stats one. It should be Prometheus specific prefix.

danbogos:
Thanks!

PR #4757, opened 2025-05-08
Fix GORM pagination with zero limit/offset + GetCDRs refactoring

ionutboangiu:
Previously, q.Limit(0) was always set, causing GORM to return no records since limit=0 was interpreted as "get 0 records" rather than "no limit". Now we only set limit and offset when they're > 0, which makes pagination behave correctly with default values.


danbogos:
Thanks!

PR #4774, opened 2025-05-21
Fix context lifecycle in scheduled actions

ionutboangiu:
Remove ctx field from scheduledActs struct and create a fresh context when actions execute via cron. This prevents "context canceled" errors that occurred when stored contexts from API calls were used for delayed execution. The context is now properly received from the caller in case of "*asap" actions.


danbogos:
Thanks!

PR #4777, opened 2025-05-23
Fix rsyslog config to capture logs across distros

ionutboangiu:
Add three conditions to properly capture CGRateS logs:

- programname startswith 'cgr-engine': catches logs from external libraries that don't use our syslogger (e.g. diameter conn errors)
- programname startswith 'CGRateS': handles deb-based distros where our syslog tag becomes the program name
- msg contains 'CGRateS': handles rpm-based distros where systemd journal overrides programname but our tag remains in message content


danbogos:
Thanks!

PR #4779, opened 2025-05-26
Close db connections in test flush helpers

ionutboangiu:
Previously, database connections would stay alive until "go test" would finish, sometimes leading to max conn errors.


danbogos:
Thanks!

PR #4795, opened 2025-06-02
Revise IPProfile definition + tests

ionutboangiu:


apis/loaders_it_test.go
```diff
@@ -996,11 +1000,54 @@ func testLoadersGetIPProfiles(t *testing.T) {
...
 				},
 			},
-			TTL:         time.Second,
-			Type:        "ipv4",
-			AddressPool: "172.16.1.1/24",
-			Allocation:  "alloc_success",
-			Stored:      true,
+			TTL:    time.Second,
+			Stored: true,
+			Pools: []*utils.Pool{
```
danbogos, line 1005: Maybe rename utils.Pool to utils.IPPool so we can know where it is coming from?

PR #4833, opened 2025-07-04
Add non-indexed variants of indexable filter types

armirveliaj:


danbogos:
Thanks, do we already have tests for matching *nistring ...?

PR #4852, opened 2025-07-18
prometheus: add node_id label to cache metrics

ionutboangiu:


engine/caches.go
```diff
@@ -443,6 +443,22 @@ func (chS *CacheS) V1GetCacheStats(ctx *context.Context, args *utils.AttrCacheID
 	return
 }
 
+type CacheStatsWithMetadata struct {
+	CacheStatistics map[string]*ltcache.CacheStats
```
danbogos, line 447: Have the inclination towards Stats here instead of CacheStatistics. Maybe we need to discuss about it.

PR #4863, opened 2025-07-30
Add ReplyState tracking with RSC/RFC stats metrics

ionutboangiu:


docs/stats.rst
```diff
@@ -134,6 +134,18 @@ Following metrics are implemented:
...
+
+\*lowest
+	Generic metric to return the lowest value of a specific field within *Events*. Format: <*\*lowest#FieldName*>.
+
+\*rsc
+	Reply success count. Counts requests where ReplyState equals "OK". Uses *ReplyState* field in the *Event*.
+
+\*rfc
+	Reply fail count. Counts requests where ReplyState is not "OK". Uses *ReplyState* field in the *Event*.
+
```
danbogos, line 148: Please rename to *repsc and *repfc. Also add *repsac (reply success average count) and *repstc (reply success total count). Same average and totals also for failed ones (*repfac, *repftc).
It will be great to find a way to add multiple stats of failed, ie: *repfc#ERR_INITIATE, *repfc#ERR_UPDATE. Same for the other failed related.

PR #4900, opened 2025-09-18
config: add sanity check for undefined prometheus conns

ionutboangiu:
previously existed only for StatSConns, added also for CoreSConns and CacheSConns


danbogos:
Thanks!

PR #4922, opened 2025-10-28
fix DPR to contain sender identity not recipient

ionutboangiu:


danbogos:
Thanks! Please do not forget to move it to 1.0 as well.

PR #4939, opened 2025-11-10
Use DH_VERBOSE only in dh_auto_build override for go version & env output

sebastic:
Setting `DH_VERBOSE` globally makes the output of the other targets too noisy.


danbogos:
Thanks!

sebastic:
Do you need separate PRs for the other branches, since you don't have the backports bot?

PR #4941, opened 2025-11-17
Use DH_VERBOSE only in dh_auto_build override for go version & env output (1.0)

sebastic:
Same changes as #4939 for 1.0 branch.


danbogos:
Thanks!

PR #5109, opened 2026-04-03
Replaced subscribing each channel at StasisStart with subscribeAll parameter at start

gezimbll:


danbogos:
This needs to be applied per configuration. Old behavior should be maintained and also only channel events we are interested about should be filtered out.

danbogos:
Thanks, let's try another approach. Please make SubscribeAll default (no config) and no subscribe per channel at all (even if coming via Stasis). Will everything work as up to now? I hope we will have the same behavior as with FreeSWITCH, where park is used for authorization and we can do accounting always since we get the events by simply listening.

PR #5112, opened 2026-04-07
resources: replace dirty flag with Stored and drop config global

ionutboangiu:
make naming consistent with IPs


danbogos:
Thanks!

PR #5121, opened 2026-04-14
fix JSON_UNQUOTE wrapping in FilterToSQLQuery 

ionutboangiu:
JSON_UNQUOTE was wrapping the entire expression (wiht comparison) instead of just JSON_VALUE. MariaDB allows it, but MySQL doesn't.


danbogos:
Thanks!

PR #5125, opened 2026-04-17
fix service state race

ionutboangiu:
shutdown could panic if it ran while a service was still starting.

also fixes shdWg leaks when Start fails and ctrl+c hanging during waits.


danbogos:
Please review failing tests.

danbogos:
Thanks!

PR #5127, opened 2026-04-20
Add AttributeIDs to ThresholdProfile to alter events triggering Threshold

gezimbll:


danbogos:
Please check tests and also try to avoid moving the Attributes to utils.

PR #5171, opened 2026-05-12
ansible: clone cgrates directly in deb_packages

ionutboangiu:
since everything is done in chroots, there is no need to build the binaries


danbogos:
Thanks!

PR #5188, opened 2026-05-21
ansible: only scp debs to pkg server in deb_packages

ionutboangiu:
the pkg play only moves the main debs out of /tmp/, the rest just kept gathering there


danbogos:
Thanks!

PR #5192, opened 2026-05-21
sessions: query sync clients in parallel

ionutboangiu:
Query bidirectional clients concurrently and give each request its own context timeout. Collect replies through a buffered channel and wait for all queries before deciding which active sessions are missing.

Clients that time out or return an error are treated as reporting no active sessions.


danbogos:
Thanks!

PR #5201, opened 2026-05-28
kamagent: respect reply_timeout when listing dialogs

ionutboangiu:
V1GetActiveSessionIDs waited a hardcoded 5s for the dlg.list reply, ignoring any reply_timeout above 5s. The sync then timed out and terminated all active sessions.

It now waits for reply_timeout. The reply channel is also buffered and drained each sync so a late reply can't block the handler or be counted in in the next run.


danbogos:
Thanks!

PR #5210, opened 2026-06-09
ees: drop the go-elasticsearch typed client dependency

ionutboangiu:
go-elasticsearch pulls in 700+ packages just to send one put.
switched to elastic-transport directly.


danbogos:
Thanks!

PR #5218, opened 2026-06-10
radagent: test charging with action provisioning

ionutboangiu:
same plan as the existing test, but the data balances come from a *setBalance action instead of being declared in the account


danbogos:
Thanks!

PR #5321, opened 2026-08-27
sessions: don't use results from failed debits

ionutboangiu:


sessions/apis.go
```diff
@@ -1413,6 +1413,7 @@ func (sS *SessionS) BiRPCv1ProcessEvent(ctx *context.Context,
 				utils.Logger.Warning(
 					fmt.Sprintf("<%s> error: %s processing event: %+v with %s for Debit",
 						utils.SessionS, err.Error(), cgrEv, utils.AccountS))
+				continue
```
danbogos, line 1416: The right approach here should be out of the if (if errors), as we do for all other functions. A continue will skip the rest of the steps from bellow and plus I think we might ignore it pretty easy.

danbogos:
Please re-check this fix after my latest re-write since I think this issue should be fixed. If not, please adapt and fix conflicts.

ionutboangiu:
The rewrite fixes this. Closing.

PR #5322, opened 2026-08-27
sessions: use session ID for removal

ionutboangiu:
ProcessEvent sessions may not have *cgrID, so use Session.ID when removing them.


danbogos:
Thanks!

PR #5337, opened 2026-09-09
ers: acknowledge JetStream messages, and settle on the processing result

moeen-ict:
### The bug

`ers/nats.go` configures its JetStream consumer with `AckPolicy: AckAllPolicy` but never calls `Ack()` or `Nak()`. Nothing is acknowledged, so the server redelivers every message after each `AckWait`, indefinitely, and any consumer of those events counts each redelivery again.

Observed on a stream with ~580 retained messages on one subject: a StatQueue fed by the reader grew by ~580 on every redelivery cycle, without bound. No `DeliverPolicy` is set either, so the first connect replays the whole subject history — fine once, unbounded when nothing is ever acked.

### Why the acknowledgement cannot live in the reader

`processMessage` hands the event to the events channel and returns, so its return value describes **parsing**, not processing — the RPC that can actually fail runs later, in `ERService`'s drain loop. Acknowledging inside the reader acks on receipt.

This is not theoretical: with an ack added inside the reader, 580 events that all failed downstream were acked anyway, `redelivered 0` under `MaxDeliver 3`.

So `erEvent` carries an `onDone func(error)`; `processMessage` puts it on the channel and reports via a `queued` bool whether it got that far; the drain loop invokes it with the result of `processEvent`. It settles *before* `exportRawEvent`, since an export failure is a side channel and should not redeliver an event that was processed.

### AckExplicit is part of the fix, not a cleanup

Under `AckAll`, acking sequence N acknowledges every message up to N. Handlers run concurrently and finish out of order, so one late success would silently acknowledge earlier failures. **Adding `Ack()` alone would trade duplicate delivery for silent loss.**

`MaxDeliver` is bounded at 3 rather than left unlimited: many reader errors are terminal (no profile matches the event), so an unbounded `Nak` loop is no better than never acknowledging.

### The subtle case

`processMessage` has five exits and each must settle exactly once. A filter that legitimately **rejects** a message returns no error and queues nothing — so keying settlement off the error alone leaves it unacknowledged and redelivering until `MaxDeliver`. Hence the explicit `queued` bool: a rejected message was consumed on purpose and is acked.

| exit | `queued` | `err` | settlement |
|---|---|---|---|
| decode error | false | err | `Nak` |
| filter error | false | err | `Nak` |
| filter rejects | false | nil | **`Ack`** |
| `SetFields` error | false | err | `Nak` |
| handed to the channel | true | nil | drain loop, on `processEvent` |

### Measured, on a live stream

```
failing path     delivered 1740 = 580 x 3   redelivered 580   ack_floor 0     then flat
succeeding path  delivered 580              redelivered 0     ack_floor 580   ack_pending 0
```

In the succeeding run `ack_floor` was observed progressing *behind* `delivered` mid-run (`ack_floor 48`, `ack_pending 514`, `delivered 580`) — acknowledgement following processing rather than receipt. A plateau alone would not show this; a stalled reader plateaus identically.

`go build ./...`, `go vet ./ers/` and `go test ./ers/` pass. The core NATS (non-JetStream) path passes a nil callback — a plain subscription has nothing to acknowledge.

### Open question for maintainers

Whether "no profile matched the event" *should* be retryable. This treats it as retryable, so such an event is delivered `MaxDeliver` times rather than once — bounded, but 3x the work where unmatched events are common. Distinguishing "no match" from "processing failed" at the drain loop would avoid that, and is a larger change than this one.


danbogos:
Hi @moeen-ict 

Unfortunately your PR is missing Contributors signing as well as corresponding unit as well as integration tests attached. Please reopen if you manage to implement these. Thanks!

PR #5341, opened 2026-09-11
upaded kamailio agent handling requests/reply through templates

gezimbll:


danbogos:
Please fix vet.

danbogos:
Thanks!

